PR #36, opened 2015-06-01
User must be logged in to create an idea

Mctalian:
Whether or not we want this for the demo (it's already rather late for
that) is up for debate I'm sure.


src/navigation/sidenav.js
```diff
@@ -26,6 +27,10 @@ angular.module('flintAndSteel')
 				}
 			};
 
+			$scope.isUserLoggedIn = function() {
```
YashdalfTheGray: You can just say `$scope.isUserLoggedIn = loginSvc.isUserLoggedIn;` here

Mctalian: Very true.

YashdalfTheGray: :+1:

src/navigation/sidenav.js
```diff
@@ -16,7 +17,12 @@ angular.module('flintAndSteel')
 
 			$scope.navTo = function navTo(state) {
 				if (state === 'idea') {
-					$state.go('idea', {ideaId: 'mock_idea'});
+					if (loginSvc.isUserLoggedIn()) {
```
YashdalfTheGray: I've actually deprecated the `state === 'idea'` part of `navTo` since there are no more home and idea buttons on the sidenav anymore. You should add this logic to another if case that checks for `state === `addidea`. I cleaned up the sidenav a little and I should have seen it when I initially commented.

YashdalfTheGray:
I think, this will need an merge from the main repo to yours using `git pull` first and then a PR from yours to the main. I just tried merging this and git can't automatically do it. We can also take this opportunity to clean up some of the logic, post final presentation!

YashdalfTheGray:
By the way, I also gave you push access. Don't know why I didn't earlier but you're welcome to work off of your own repo and make PRs still. That's how it should be done, ideally anyway!

Mctalian:
Yeah we'll leave this one out for now. I was limited in my abilities at work so I was hoping a PR to me, then a PR back to you would work, but alas, it just made things worse :/

PR #37, opened 2015-06-01
Config Script to generate canned ideas before Demo

Mctalian:
Instead of stressing about creating these ideas the day of the demo, we
can enter them in the config.js script and they will be there when we
get set up.


Mctalian:
This can be temporary until the demo is over.

YashdalfTheGray:
Should we do this pre-running the server or we can make a gulp task that just does this. Something like `gulp generate:data`? The only reason I say this is that we're using gulp as the runner/tester/linter tool rather than npm.

LongArmMcGee:
the switch to gulp implementing the script sounds more appropriate, and this may be a semi-permanent script until we get to a more stable code base?

Mctalian:
I agree. I did not think of that. I can make the modifications for this one
later today.

YashdalfTheGray:
Done and then I can merge it.

README.md
```diff
@@ -4,39 +4,45 @@
 
 After cloning the git repo run `npm install` from a terminal window.
 
-**Note**: On unix based platforms, this will need to be run with `sudo` and on Windows, the command prompt will require administrative privileges. 
+You will likely also have to run `npm install gulp-cli -g` in order to run gulp. This command installs `gulp` globally so it can be used in the command line.
```
YashdalfTheGray: I looked into this and turns out that it's better to have gulp installed globally rather than just gulp-cli. You want to change that?

package.json
```diff
@@ -5,7 +5,7 @@
   "main": "index.html",
   "scripts": {
     "test": "echo \"Error: no test specified\" && exit 1",
-    "postinstall": "bower install",
+    "postinstall": "bower install && node config",
```
YashdalfTheGray: Since we have the gulp task, we don't need this, do we?

README.md
```diff
@@ -4,39 +4,45 @@
...
+Running `gulp jshint` will lint any JavaScript file under the `src` folder excluding `src/lib`.
 
 ## Clearing Persistent storage
 
-The express server app uses [docstore](https://www.npmjs.com/package/docstore) to store users and ideas in JSON format. If this storage location gets messy, `gulp clean:db` can be used to clear up the datastore. 
+The express server app uses [docstore](https://www.npmjs.com/package/docstore) to store users and ideas in JSON format. If this storage location gets messy, `gulp clean:db` can be used to clear up the datastore.
+
+## Generating Canned Idea datastore
+
+To start with some predefined data, `gulp generate:data` populates the datastore with pre-populated ideas.
```
YashdalfTheGray: Thanks for the spelling error corrections and also, it seems like your text editor doesn't like spaces at the end of lines. I wonder if that's going to cause layout issues with markdown or something, it shouldn't.

README.md
Mctalian: This might be different than grunt then. I know grunt if you install it globally, it doesn't work, but grunt-cli will. I will change this if you're sure.

package.json
Mctalian: Correct, my mistake. This actually won't work anymore and would probably cause errors. Good catch.

README.md
Mctalian: Yeah, it's a very picky text editor. But I use it for a personal project and the markdown comes out just fine.

YashdalfTheGray: Perfect!

PR #38, opened 2015-06-03
Experiment real time experience

Mctalian:
This was an experiment to simulate realtime data for the sidenav and for the ideaView. It turned out pretty nicely. Any new idea that is created shows up on all sessions near instantaneously. Any new likes, comments, and backs also show up in the idea view. Now if only I could animate them :stuck_out_tongue_closed_eyes:


YashdalfTheGray:
I think we can leave animation for after final presentation and angular and ngMaterial have pre-canned animations that we can use, after doing some research.

PR #41, opened 2015-06-03
Continous Integration using CircleCI

Mctalian:
Adds some info on how someone can set this up pretty easily. If someone tries it and notices the instructions are a little off, please update them.


YashdalfTheGray:
I'd like to use Travis, it seems to be the standard in the JS/HTML realms. Thoughts?

Mctalian:
Ah, yes. This could work for flintandsteel because it's a public repo. When I was looking for CI options on my personal project, I thought I had to pay for Travis for private repos.

I have experience with Circle, but Travis would fit our needs as well, at it does seem like the standard.

Issue #46, opened 2015-06-06
IE Compatibility


Mctalian:
We don't have to fix it... just so we're all aware.

YashdalfTheGray:
What lapse in judgement caused opening up the website in IE in the first place?

Mctalian:
Update: Since PR #70, there are 2 known issues with IE, the rest have been fixed.
* The home page, all of the columns at the bottom like "Innovate, Collaborate, Achieve" all get mashed into the middle of the screen.
* Each idea page doesn't seem to show the md-tabs of comments and backs.

I firmly believe these can be fixed. I can look into these later.

Mctalian:
The homepage can be "fixed" by removing `center` from `md-content`'s `layout-align` at the top of `homeView.tpl.html`. This makes the Rok-Starter logo super disproportionate, so there's still some work to do to make the homepage work perfectly for IE. (This change may also make the logo disproportionate on Chrome as well, haven't tested)

Mctalian:
As for the tabs: Our exact Idea Page (I pasted the md-content portion into a basic angular app Pen) source code shows up in CodePen in IE... So my leading theory is theme issues, but I'm not entirely sold on that.

Mctalian:
The tabs work fine with version 0.9.4 of angular-material.min.css

I've opened an issue on angular-material.

Mctalian:
Tabs
https://github.com/angular/material/issues/3742
https://github.com/angular/angular.js/issues/12348

Mctalian:
Reopening to keep an eye on the Material Design / Angular Versions to ensure that things are fixed before upgrading to a new Material Design or Angular version

PR #70, opened 2015-07-06
IE fix attempts

Mctalian:
This makes IE a little more usable. You can login and see ideas. There's still some work to be done.

Comment and Back tabs are still not visible on idea pages.

Front page still has all the content squished together.


Mctalian:
#46

PR #82, opened 2015-09-11
Travis-CI integration!

YashdalfTheGray:
We're going to have to add the little travis CI status indicator once this is on master. 

This will run tests every time anything is committed to the repo through commits and through PRs, this way we can make sure code coming in is actually good!

Requested reviewers - @Mctalian @davethenipper

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ node_js:
   
 before_script:
   - npm install -g gulp
+  - npm install -g bower
   
 script: 
   - gulp test:client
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -0,0 +1,12 @@
+language: node_js
+node_js:
+  - "0.10"
```
Mctalian: I would recommend adding a subminor here as well. Easier to reproduce builds. Similar thought to our bower and package.json. Keeps any changes in libraries from breaking our app as well.

Mctalian:
:+1 Approved.

davethenipper:
Approved

YashdalfTheGray:
Added a patch version to node, merging now.

PR #84, opened 2015-09-16
Idea Editing and Removes Author from AddIdeaView

Mctalian:
Angular-* upgraded to 1.4.4
Angular-Material upgraded to 0.11.0
This makes md-fab-toolbar useable. Also might knock out a few of the 3rd party issues.

Ideas can now be edited by the author.

Consequently, a new idea can no longer provide the author of the idea. It is automatically their name.

Requested Reviewers: @YashdalfTheGray, @davethenipper


src/ideas/ideasView/ideasView.tpl.html
```diff
@@ -1,10 +1,46 @@
...
+	      </md-button>
+	    </md-fab-trigger>
+	    <md-toolbar>
+	      <md-fab-actions class="md-toolbar-tools">
+	        <md-button aria-label="edit" class="md-icon-button" ng-click="ideasView.enableEdit = true;">
+	          <md-icon md-svg-icon="editor:mode_edit"></md-icon>
+	        </md-button>
+	      </md-fab-actions>
+	    </md-toolbar>
+	  </md-fab-toolbar>
```
YashdalfTheGray: How well does the fab toolbar work?

YashdalfTheGray:
Approved.

src/ideas/ideasView/ideasView.tpl.html
Mctalian: Pretty well. I was also going to add the delete option to it later. We can do the whole UX evaluation in a demo sometime within the next week.

Mctalian:
Added a few more commits for getting rid of most of the 3rd party bugs. I didn't branch like I normally do, so that's my bad.

davethenipper:
Approved

Mctalian:
Md-fab-toolbar doesn't seem to work quite right with IE. Will submit a defect to material tomorrow.

In the meantime it will remain and I'll temporarily fix it when I add deleting ideas.

PR #92, opened 2015-10-01
Delete comments

Mctalian:
Comments are not "deleted" but the text is replaced with "This comment was deleted", the timestamp is updated, and the comment is shown with a line through it. Braun and I played around with a few different options and we decided on this one.

Reviewers: @bcbrennecke , @davethenipper


davethenipper:
Approved

bcbrennecke:
Approval

PR #107, opened 2015-10-27
Graphics Updates

bcbrennecke:
Updated the Graphics through the application to the new logos.
This is not IE compatible.  Passing to @Mctalian 
Review me: @YashdalfTheGray , @davethenipper


src/homeView/homeView.tpl.html
```diff
@@ -1,28 +1,21 @@
 <md-content layout="column" layout-align="start center">
-    <p></p>
-    <p></p>
-
-    <!--<font size="12"><b>Rok-Starter</b></font>-->
-    <img src="assets/logo2.png">
-
-    <!--<img src="assets/fire.png"/>-->
+    <img src="assets/InnovationChallengeLogo.png" width="43%">
```
Mctalian: if you add `height="28%" this will work on IE

Mctalian:
Did you forget to commit `index.html` where you used the Lightbulb in place of "Rok-Starter"?

Mctalian:
The rest of the IE stuff is annoying... we might have to open an issue in Material Design to see if they can figure out what is going on.

Mctalian:
I actually have everything working but mobile friendly on IE. Will anyone ever actually use IE mobile or with that small of a screen? hmmm.

davethenipper:
They probably will.  But should they? The answer is no.

Mctalian:
I find that use case to be in the 5%. We should check it on iPhone browsers though. The whole site, not just this page.

Either way, I sent the IE updates to @bcbrennecke in an email so expect those updates soon-ish.

YashdalfTheGray:
@davethenipper, the answer to all of IE should be a no. I know it's the company supported browser and all but I'm fairly certain that our target audience is going to using chrome. Having the support is good, I don't think we should focus on it that much. Just my thoughts.

Mctalian:
I agree for IE mobile or small screens. That's very unlikely considering
the company provides iPhone not windows phone. IE desktop, I feel, should
be something we still care about. I don't mind doing the needed updates

YashdalfTheGray:
Approved.

davethenipper:
Approved!

Mctalian:
Approved :+1:

PR #110, opened 2015-10-28
Replacing docstore with mongo

YashdalfTheGray:
This includes changes to the server backend to use mongo. 

Reviewers - @Mctalian @bcbrennecke @davethenipper


server/server.js
```diff
@@ -168,51 +166,47 @@ app.post('/login', function handleAuthentication(req, res, next) {
...
-                            return res.status(200).json({
-                                status: 'AUTH_OK',
-                                id: doc.accountId,
-                                username: doc.key,
-                                email: doc.email,
-                                name: doc.full,
-                                likedIdeas: doc.likedIdeas
-                            });
+                else {
+                    db.collection('users').insertOne({
```
Mctalian: So it looks like every time someone logs in, it will create a new user document in the users collection. Can we look to see if there is a document with that username already present when they login and overwrite their data if they are? And if we can't find them, insert a new doc?

server/server.js
```diff
@@ -281,25 +275,22 @@ app.post('/deleteidea', function(req, res) {
...
-            userDb.save(doc, function(err, doc) {
-                if (err) {
-                    console.log(chalk.bgRed(err));
-                }
-                else {
-                    console.log(chalk.bgGreen('Document with key %s updated in account.'), doc.key);
-                    res.sendStatus(200);
-                }
-            });
+    // This will probably become an every login thing with LDAP anyway.
```
Mctalian: Yes, this is how I had it for docstore, and I think it makes sense to do this. I don't see the data changing very often but the only reason we'd care is if they login and info was slightly different, then we'd update it. We may have to adjust the login route logic to reflect that same thing. (See my line comment above in the login route).

server/ideas.js
```diff
@@ -1,48 +1,19 @@
...
-    }
-    else {
-        userDb = store;
-    }
-});
-
 exports.create = function(id, title, description, author, likes, comments, backs, cb) {
     "use strict";
 
     var idea = Idea.create(id, title, description, author, likes, comments, backs);
```
Mctalian: I'm going to make an issue to add team info to the Idea schema. Just putting a comment here to remind myself.

I know @davethenipper wants the author added to the team by default, so I can do that in there.

server/server.js
```diff
@@ -6,53 +6,51 @@ var express = require('express'),
...
+        if (errIdea) {
+            console.log(errIdea);
+        }
+        else {
+            db.createCollection('users', function(errUsers, collectionUsers) {
+                if (errUsers) {
+                    console.log(errUsers);
+                }
+                else {
+                    collectionUsers.insert({ "myUser": "Test Testerson III" });
```
Mctalian: Is this test code/needed?

gulpfile.js
```diff
@@ -74,7 +116,7 @@ gulp.task('start:dev', ['_cleanUp', 'test:client', 'inject'], function() {
     });
 });
 
-gulp.task('start:prod', ['_cleanUp', 'test:client', 'inject'], function() {
+gulp.task('start:prod', ['test:client', 'inject'], function() {
```
Mctalian: I assume we can't add mongo:start to the task dependencies because there's no way to determine if the server is already running when we call this?

package.json
```diff
@@ -49,6 +49,8 @@
     "karma-cli": "0.0.4",
     "karma-jasmine": "0.3.6",
     "karma-phantomjs-launcher": "0.1.4",
+    "mkdirs": "0.0.2",
+    "mongodb": "^2.0.45",
```
Mctalian: Can we make this 2.0.45 so it doesn't auto upgrade on us?

server/server.js
```diff
@@ -168,51 +166,47 @@ app.post('/login', function handleAuthentication(req, res, next) {
                 });
             }
 
-            req.login(user, function(err) {
```
Mctalian: Wait wait wait... this is needed- it's a passport call to authenticate the user with our site.

server/server.js
```diff
@@ -6,53 +6,51 @@ var express = require('express'),
...
-    datastore = require('docstore'),
     bodyParser = require('body-parser'),
     external = require('external-ip')(),
     passport = require('passport'),
     WindowsStrategy = require('passport-windowsauth'),
     ip = require('ip'),
-    _ = require('lodash'),
-    ideas = require('./ideas'),
-    ldapAuth = require('./secrets/ldapAuth');
+    ldapAuth = require('./secrets/ldapAuth'),
```
Mctalian: I'm going to put in an issue where this is only required of people trying to gulp start:prod

package.json
```diff
@@ -49,6 +49,8 @@
...
     "passport": "0.1.15",
     "passport-windowsauth": "*"
```
Mctalian: Unfortunately this passport-windowsauth seems dead. But RA has moved away from LDAP and is using ADFS instead. We need to talk to IT about that again, but we discussed it at one point.

gulpfile.js
YashdalfTheGray: That's how I had it to start but when I started testing, I realized that `mongod` (mongo's server daemon) would be killed when there were build errors in the server or anything went belly up even though it wasn't mongo's fault. That's just how gulp handles things I guess so I took that part out.

The new workflow is running `gulp mongo:start` in one terminal window, `gulp start:[dev|prod]` in another. Once you're done, Ctrl+C the `gulp start` command and use that terminal to run `gulp mongo:stop`. I think that it's a little more logical that way and makes sure that `mongod` isn't even killed because of errors that don't concern it.

package.json
YashdalfTheGray: Yup, that makes sense. I'll do that tonight.

server/server.js
YashdalfTheGray: I tried that, it gave me a ton of errors. This way, it solely depends on what you put in your ldapAuth.js file. Plus, if you make this an optional dependency, you're going to need to check if `ldapAuth !== undefined` everywhere else and I don't think that's worth it.

package.json
YashdalfTheGray: Ooh, this might be better. I think ADFS supports encryption and also it *may* support just http calls, which is good for us.

server/server.js
YashdalfTheGray: Oh I see, this helps me make sense of something else in the code that I couldn't quite place. Will update.

server/server.js
YashdalfTheGray: The creating collections is probably needed, I need to test that but I'll remove the insert statement.

server/ideas.js
YashdalfTheGray: I'll add the fact that the author needs to go on the team to the json schema then.

server/server.js
Mctalian: Sorry, I mentioned this because I was going to look into it later, nothing for you to worry about here. There's an issue up and I'll investigate it, but it may end up that it has to stay this way.

YashdalfTheGray:
I fully expect some of this stuff to break and crash but that's why I created the cleanup task (issue #113).

gulpfile.js
Mctalian: :+1:

Mctalian:
Yeah, that's fine. I'll go through and test some things once this gets merged and I can help out with clean up as well.

Approved

davethenipper:
Approved!

PR #116, opened 2015-10-30
Fix dev login

Mctalian:
Password was not being decoded and the fake credentials weren't short enough for us complainers.

@LongArmMcGee @bcbrennecke


LongArmMcGee:
Approve

bcbrennecke:
Sureee

PR #120, opened 2015-11-01
Added About page

bcbrennecke:
Initial commit for an about page.  This information will be dynamic as new developers make commits and the site evolves. @YashdalfTheGray @Mctalian


bcbrennecke:
Approve me!  @YashdalfTheGray @Mctalian

YashdalfTheGray:
This looks good. Approved.

src/assets/app.css
```diff
@@ -98,6 +98,26 @@ body {
...
\ No newline at end of file
+}
+
+.about-technologies{
+    padding-top: 15px;
+    padding-left: 15px;
+    padding-right: 15px;
+    padding-bottom: 15px;
+}
+.about-padding-bottom{
```
LongArmMcGee: I don't like this page-specific naming convention. Is this a common practice?

YashdalfTheGray: Yeah, it's common practice. Usually you would start with something specific and if you find yourself using the same CSS again somewhere, you would make the class more generic. The specificity is to avoid class collisions.

If we were using [SASS](http://sass-lang.com/guide) or [LESS](http://lesscss.org/features/), we'd be able to create classes and inherit from them and all but we're using plain CSS so we don't get those fancy features. It works for this project though so there isn't a reason to change that.

Mctalian:
Approved. We might want to look into dressing this up to make it pretty later, but it definitely gets the job done for now.

src/assets/app.css
bcbrennecke: I used this because I saw that we had gone down the road of making very specific css classes and wanted to start some sort of little organization to them.  Always glad when a random idea lines up with common practice

PR #128, opened 2015-11-04
Fix nodemon

### LINKED ISSUE #126: Nodemon is watching the DB!!!
Anytime a write occurs to the DB, the server restarts. No bueno. It's a simple fix, and I did so on the production server as a stop gap, but we'll need to throw an official PR up.

In case anyone wants to take care of it, you change `'spec/*'` in the `'ignore'` field to `'server\datastore\*'` for both Prod and Dev starts. It should look like the following.

```
gulp.task('start:dev', ['_cleanUp', 'test:client', 'inject'], function() {
    "use strict";

    nodemon({
        script: 'server/server.js',
        env: { 'NODE_ENV': 'development' },
        'ignore': ['server/datastore/*']
    });
});

gulp.task('start:prod', ['_cleanUp', 'test:client', 'inject'], function() {
    "use strict";

    nodemon({
        script: 'server/server.js',
        env: { 'NODE_ENV': 'production' },
        'ignore': ['server/datastore/*']
    });
});
```

Mctalian:
Closes #126 

@YashdalfTheGray @LongArmMcGee

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -112,7 +112,7 @@ gulp.task('start:dev', ['test:client', 'inject'], function() {
     nodemon({
         script: 'server/server.js',
         env: { 'NODE_ENV': 'development' },
-        'ignore': ['spec/*']
+        'ignore': ['server/datastore/*']
     });
 });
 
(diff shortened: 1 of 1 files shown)


YashdalfTheGray:
Good call on this, missed that. Approved.

davethenipper:
Approved!

Issue #105, opened 2015-10-26
Login "screen"


monotkate:
When you click to login it's using a form that looks like it's designed to be a drop down, but isn't a drop down.  Almost the entire page ends up white and empty with the tiny white login box and it looks disjointed.  Possible solution is to have as a drop down that overlays main page, or to tweak the design to look better on a full page.

Mctalian:
That's a good point. I've thought about this myself. I'm thinking we're
going to do most of the UX improvements after we've got a lot of the main
functionality done. Unless you want to give it a shot yourself, I like the
drop down or modal pattern.

monotkate:
If I've got some time I'll poke around at it

PR #133, opened 2015-11-12
Login switched to Dialog

monotkate:
Changed the way login works so it uses a dialog over the main page instead of navigating to a separate page.

Combined the loginView controller code into the toolbar controller for the dialog.  Removed loginView.js because it was no longer used.


monotkate:
Closes #105

LongArmMcGee:
Hawt

monotkate:
@YashdalfTheGray @Mctalian

src/index.html
```diff
@@ -57,10 +56,21 @@
             <h2><span><img ui-sref="home" class="center-img" src="assets/Innovation_Challenge_Bulb.png" width="25"/></span></h2>
             <span class="padding-left-20 idea-likes-click" ui-sref="ideabrowse">All Ideas</span>
             <span flex></span>
-            <md-button class="md-icon-button" aria-label="Account" ng-click="accountClick()">
-                <md-icon md-svg-icon="action:perm_identity" ng-if="!isUserLoggedIn()"></md-icon>
-                <identicon username='username' size='24' ng-if="isUserLoggedIn()" ng-init="getUsername()" class="identicon"></identicon>                
-            </md-button>
+
+            <div class="md-padding dialogdemoBasicUsage" id="popupContainer" ng-cloak="">
```
YashdalfTheGray: In this code block, you're using 2 spaces as indentation, change it up to 4 to be consistent.

src/navigation/toolbar.js
```diff
@@ -38,7 +66,68 @@ angular.module('flintAndSteel')
...
+                    controller: DialogController,
+                    templateUrl: 'users/loginView/loginView.tpl.html',
+                    parent: angular.element(document.body),
+                    targetEvent: ev,
+                    clickOutsideToClose: true
+                })
+                .then(function(answer) {
+                    $scope.loginUser(answer);
+                }, function() {
+                    $scope.status = 'You cancelled the dialog.';
```
YashdalfTheGray: Canceled

src/index.html
YashdalfTheGray: Also, where is the `dialogdemoBasicUsage` class defined?

src/users/loginView/loginView.tpl.html
```diff
@@ -1,29 +1,41 @@
...
+            </md-input-container>
+            <br></br>
+            <md-dialog-actions flex layout="column">
+                <md-button class="md-raised md-primary login-button" ng-click="login(account)" flex>
+                    Login
+                </md-button>
+            </md-dialog-actions>
+            <p></p>
+        </md-dialog-content>
+        <!--<md-dialog-actions>
```
YashdalfTheGray: You can just remove this commented out section.

monotkate:
@YashdalfTheGray Should have all your comments fixed now

YashdalfTheGray:
Approved. Waiting on one more.

Mctalian:
I checked it out locally and it looks really good. The only thing I'm a little iffy on is that we blew away an entire test spec. But taking a closer look it only had 3 cases, which rely more on the server doing its job and the last test case was no longer used because we no longer have a sign-up process.

Looks like you have some merge conflicts after I merged in my large changes.

Other than that, I approve.

PR #137, opened 2015-11-13
Clean up

Mctalian:
See commit description for changes. :beers:


Mctalian:
@alanlgirard and @bcbrennecke and Andrew watched over my shoulder as I implemented and tested these chats changes.

@YashdalfTheGrey, @monotkate one of you take a sober look to make sure we're good here.

Mctalian:
Damnit... @YashdalfTheGray

Mctalian:
Also disregard the word "chats" a few comments above. Inadvertent word.

src/ideas/addIdeaView/addIdeaView.spec.js
```diff
@@ -53,12 +53,12 @@ describe('AddIdeaViewCtrl', function() {
     it('should use the user\'s name as the author', function() {
         var idea = {
             title: 'Test Title',
-            author: 'Test',
```
Mctalian: I made a mistake here. This should remain the same.

src/ideas/addIdeaView/addIdeaView.spec.js
```diff
@@ -53,12 +53,12 @@ describe('AddIdeaViewCtrl', function() {
...
+            authorId: 3,
             description: 'This is a test idea.'
         };
         scope.addNewIdea(idea);
 
-        expect(idea.author).not.toBe("Test");
-        expect(idea.author).toBe("Darth Vader");
```
Mctalian: This line should be .toBe(1)

Or whatever the Darth Vader _id  is in the login mock.

The line above it should be undeleted and the two new lines should be disregarded.

monotkate:
Not sure it's necessary to route home if they aren't logged in for a new idea.  Maybe just through up an $mdToast message instead telling them they aren't logged in?  Should be able to login from any page.

I just tested this, you can log in from any page, but it will route you back to home after you're finished.  I'll fix this bug when I fix the lack of enter on login (I already know the fix for that)

Mctalian:
This PR fixes that as well

Mctalian:
"That" being the route back home on success. Not the enter button.

monotkate:
Alright, cool.  I stand by notifying that they aren't logged in and staying on the same page instead of routing them back to home though for a 'need to be logged in' action

Mctalian:
It's only if someone tries to go directly to the URL for adding a new idea
will it reroute to home. I do like the toast idea so maybe if someone does
try to access that url, it routes and toasts.

The "add new idea" button doesn't show up if they aren't logged in, so I
don't want to allow them to access it at all without being logged in.

monotkate:
Agreed.  That makes sense for the redirect then, still digging my way through what all of the code does.  The rest of it looks good to me, Approved

Mctalian:
Created Issue #138 to cover what we discussed here.

PR #140, opened 2015-11-13
Added OK to logout toast and enabled login on enter

monotkate:
@YashdalfTheGray @Mctalian @bcbrennecke


alanlgirard:
APPROVED

Mctalian:
Beautifully done. Approved.

Question though, how did adding `type="button"` to the Close button make enter work for the Login button? I have no doubts that it works because I just saw it first hand but I'm curious as to why it works.

monotkate:
http://stackoverflow.com/questions/17551220/angular-ui-dialog-and-form-submit-on-enter-key

Basically, it defaults to enter working, so it was entering on the cancel button.  I changed cancel to be just a button only, so then it went back to doing enter for login.

bcbrennecke:
Approve!

PR #141, opened 2015-11-14
Added menu drop down when logged in for profile navigation or logout

monotkate:
@YashdalfTheGray @Mctalian @LongArmMcGee

### DIFF
--- a/src/users/accountView/accountView.tpl.html
+++ b/src/users/accountView/accountView.tpl.html
@@ -8,5 +8,4 @@
         <h2><center>{{user.name}} (<i>{{user.username}}</i>)</center></h2>
         <h4><center>{{user.email}}</center><h4>
     </div>
-    <md-button class="md-raised md-warn" ng-click="logout()" flex>Logout</md-button>
 </md-content>
--- a/src/users/accountView/accountView.tpl.html
+++ b/src/users/accountView/accountView.tpl.html
@@ -8,4 +8,5 @@
         <h2><center>{{user.name}} (<i>{{user.username}}</i>)</center></h2>
         <h4><center>{{user.email}}</center><h4>
     </div>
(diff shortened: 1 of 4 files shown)


monotkate:
I maaaay have broken logout tests? >.>

src/users/accountView/accountView.tpl.html
```diff
@@ -8,5 +8,4 @@
         <h2><center>{{user.name}} (<i>{{user.username}}</i>)</center></h2>
         <h4><center>{{user.email}}</center><h4>
     </div>
-    <md-button class="md-raised md-warn" ng-click="logout()" flex>Logout</md-button>
```
Mctalian: Is there any harm in keeping the logout button on the account page as well as in the new menu? Just wondering.

If we opt to remove it like you're suggesting, you can delete the whole `describe('$scope.logout', ...)` section in `accountView.spec.js`.

Mctalian:
Other than that comment above, I think it looks good. Maybe add icons to each of the items in the drop down? Might look tacky, not sure.

YashdalfTheGray:
I think it would look tacky. Do text only.

Yash Kulshrestha
yash.kulshrestha@gmail.com
(614)563-1163

src/users/accountView/accountView.tpl.html
monotkate: I could leave it on the account page.  No harm, I was just figuring they could still logout from the drop down while on the account page.

YashdalfTheGray: Yeah, do that.

monotkate:
Updated putting the login button back on the account page - unbreaking tests :)

Good to go? @YashdalfTheGray @Mctalian

YashdalfTheGray:
Yeah, should be good to go.

Mctalian:
Approved! :+1:

PR #143, opened 2015-11-14
Mailto links

LongArmMcGee:
Would an html template allow us to 'fuction-ize' these mailto links? If we keep the # of mailto links <5 that probably isn't necessary


Mctalian:
You can make some sort of angular directive to functionize it. Braun did it for the items in the All Ideas page.

Mctalian:
My opinion: given the fact that you are including the mail subject in the mailto link (which I approve of, by the way), that's enough reason to wrap this in a directive.

Also we'll want to put the styling in a CSS rule of some sort. Embedded style should only be used when necessary (which is pretty rare).

src/ideas/ideasView/ideasView.tpl.html
```diff
@@ -100,7 +100,7 @@
                             <div class="md-list-item-text">
                                 <h3>{{back.text}}</h3>
                                 <p>
-                                    {{back.author.name}} <small class="extra-space-left" ng-bind="momentizeTime(back.time)"></small>
+                                    <a href="mailto:{{back.author.name}}?Subject=Innovation%20Challenge" target="_top" style="text-decoration: none; color: grey;">{{back.author.name}}</a> <small class="extra-space-left" ng-bind="momentizeTime(back.time)"> </small>
```
Mctalian: this should be `"mailto:{{back.author.mail}}..."` not `back.author.name`

Mctalian:
There's actually already a `mailto-format` class in the CSS. It's very similar to your embedded styles as well.

src/ideas/ideasView/ideasView.tpl.html
```diff
@@ -3,7 +3,7 @@
     <md-card-content>
         <div ng-hide="ideasView.enableEdit">
             <h1>{{idea.title}}</h1>
-            <h3>by {{idea.author.name}}</h3>
+            <email-link author-obj="idea.author"></email-link>
```
Mctalian: Do we want to keep the h3 size and the "by" here still? Something like:

```html
<h3>by <email-link author-obj="idea.author"></email-link></h3>
```

What do you think?

src/utilities/emailLink.js
```diff
@@ -0,0 +1,18 @@
...
+    [
+        function() {
+            "use strict";
+            
+            return {
+                restrict: 'E',
+                scope: {
+                    authorObj: '='
+                },
+                templateUrl: 'utilities/emailLink.tpl.html'
```
Mctalian: can you add `replace: true` after `templateUrl`? Don't forget the comma

LongArmMcGee: http://stackoverflow.com/questions/24194972/why-is-replace-deprecated-in-angularjs

Do we want to use replace?

Mctalian: In my opinion, yes. I'm willing to be wrong though

Mctalian: Sorry, I didn't see the link you put. If it's being deprecated, you're right. Leave replace out.

Mctalian:
Comment or change to idea author "by" syntax and size is the only thing I'm waiting on. Otherwise it's approved.

Line comment on line 6 of `src/ideas/ideasView/ideasView.tpl.html` is what I'm referring to.

PR #145, opened 2015-11-15
Large Back-end Overhaul/Refactor

Mctalian:
This PR will bring a LOT of new back-end stuff as well as stability improvements. I will do my best to outline these changes to make reviewing a bit easier because I am aware there is A LOT here. In hindsight, I could have broken these up a bit, or started the PR earlier, but I was super motivated this weekend and decided to just knock a bunch of things out that were bothering me. I think that the overall utility gained from this PR will overshadow the large effort of reviewing, but I guess you'll be the judge of that :stuck_out_tongue_closed_eyes:

## DB Schema
I made some changes to the actual `database-schema.json` file to start. The item of true importance is that it now accurately affects what we actually have going on in the product. Everything you see in that file is actually implemented in the code now. Each collection (`users`, `ideas`, `comments`, `events`) is modeled on the server side in their own folders (e.g. `server\users\userModel.js`). The actual interface is located in a similar file next to the model (e.g. `server\users\users.js`). The data generator (`generateData.js`) has been updated to utilize these files as well as the new schema for all of the collections.

## DB Ops Relocation
Speaking of the database, I moved all of the DB-related operations (the raw calls to the mongo library) to its own file. This helps makes the code DRY ("Don't Repeat Yourself") and enables us to switch out the back-end much easier in the future if we decide that MongoDB doesn't meet our needs (I don't foresee this). It ALSO helps with Connection Pooling for connections to the MongoDB server. I implemented connection pooling last week but we were still opening multiple connections (one in `server.js`, one in `ideas.js`, etc.). Moving all of the DB-related operations to its own file enabled us to open one and only one connection the MongoDB server throughout the ENTIRE app. This means that if 10 people are using the app, they are all leveraging the same MongoDB connection. This should help a LOT with scalability.

## Server Routes Relocation
Along with moving the DB ops out of the `server.js` file (and other places) I also moved the routes out of `server.js`. These are our API calls. For instance, when a user creates a new idea, the client-side application makes an HTTP POST request to our server. The route is `/idea` and is represented in Node/Express/Server-side as such:
```
app.post('/idea', function(req, res) {});
```
All of these routes can now be found in `server\routes.js`. In the future, I'd like us to make the REST API a bit more intuitive and move related routes to their own files as well. Much more emphasis on the moving related routes to their own files.

## Moved Getting By ID to Server Side
Since we have 4 separate collections, we need to be able to reference data between them. For example, `idea` objects have an `authorId`, an `eventId`, and multiple `commentIds`. In a PR last week, I made that magic happen on the client-side but I wasn't insanely happy about it. In this PR, it is moved to the server-side. This required us to pick up another dependency called `bluebird` to leverage Promises. All of that is handled in `server\replaceIds.js` and it is pretty cool stuff if you're interested. Just a bit of gloating: I wrote nearly all of that in one shot and, aside from one or two minor issues, it worked right away. That always feels really good when that happens!

Anyway, being on the server side, it is much more reliable, there were definitely some kinks in the client-side version. So now everyone should have access to whatever they need!

## Miscellaneous Clean Up
Due to the server-side changes, I had to make some client-side changes as well. Those are sprinkled throughout this PR. Most of them are actually in test specs or service mocks actually.

Minor change to the gulpfile so that any external commands we run (e.g. in `gulp mongo:start`, `gulp generate:data`, `gulp poop`, etc.) it will show their console logs without the trailing newline that is automatically appended when using `console.log()`.

`morgan` is only enabled in Production builds now. I was sick of it cluttering my console window when I was trying to debug this stuff. Can be undone if we miss it. Ideally, I'd like to log a lot of the console logs and morgan data in log files so we can actually try to diagnose crashes.

I need to do one more thing, and that's make sure the production server and LDAP stuff works with the moving of all these files. I'll do that after I submit this and the commit will be appended to the PR automagically.

**Any Questions, please ask. This is a lot of stuff but it should unblock a LOT of stuff and enable a lot of new features in the near future. I highly recommend everyone takes a look at this because there are a lot of great learning opportunities here, if nothing else.**

@YashdalfTheGray, @davethenipper, @bcbrennecke, @LongArmMcGee, @monotkate, @alanlgirard, @Daviddt17 

I think that's everyone and everything.


server/users/userModel.js
```diff
@@ -0,0 +1,52 @@
...
+    this.email = email;
+    this.nickname = nickname;
+    this.title = title;
+    this.likedIdeas = [];
+
+    return this;
+}
+
+// NOTE: remember to pass in "user._json", not just "user"
+function UserLDAP(ldapObj) {
```
YashdalfTheGray: I don't know how the ldap object is laid out but you could check for if user has a `._json` property and use that if someone does make the mistake of passing in user or (this is what I prefer), solidify the contract to throw and exception when `user._json` is defined and throw it back to the user.

server/routes.js
```diff
@@ -0,0 +1,311 @@
...
+
+        IdeasInstance.on("newHeaders", updateHeaders);
+
+        req.on("close", function() {
+            IdeasInstance.removeListener("newHeaders", updateHeaders);
+        });
+    });
+
+    return module;
+};
```
YashdalfTheGray: I've been meaning to look at CRUD for our application and I think this kinda paves the path for us to use CRUD.

server/users/userModel.js
Mctalian: Here's an example of the LDAP object. Rockwell's is slightly different, I believe, as far as some of the field names, but the idea is the same. https://github.com/auth0/passport-windowsauth#example-profile-from-ldap

As for `user` vs `user._json`, it might actually be better to just pass in user and handle it then. Since this function calls out that it's expecting the LDAP object in the function name.

server/routes.js
Mctalian: I was thinking the same thing! Then we can have multiple route files as well for each type of route and start using `$resource` on the client side.

server/users/userModel.js
YashdalfTheGray: Yeah, you're right. I was mainly just wondering if the `_json` object had another nested `_json` object in it or not but it seems like it doesn't. I would still throw an exception or fail some kind of gracefully when passed an illegal object.

YashdalfTheGray:
I want someone to run the current master under OSX and see if they get the `internal watch failed, watch EMFILE` error. @bcbrennecke, wanna give it a shot?

Mctalian:
So as far as that watch error, do you only get it on the current master and not on this PR?

davethenipper:
Approved by the team!

davethenipper:
Pending Yash's merge and approval

YashdalfTheGray:
I'm getting the `EMFILE` issue with master. I haven't pulled your stuff yet. Since you split it up, I was thinking that that might have changed it.

YashdalfTheGray:
Also, approved.

Mctalian:
I'll merge after you merge if you approve. We will have a conflict so I'll
resolve it on my end. Unless you can't get to yours tonight/tomorrow then
I'll just merge.

YashdalfTheGray:
Yeah, I just edited mine. And either way works with me.

PR #148, opened 2015-11-17
Team Editing/Colored Chips

alanlgirard:
-Added chip colors based on backing type (still need to figure out custom md-chip-templates for the backing selection)
-Added ease of navigation for editing team by author
-Cleaned up Backers/Team layout so it will display properly on smaller screens
-Cleaned up whitespace/newlines on toastSvc.js and accountView.js

@Mctalian @davethenipper @monotkate


src/assets/app.css
```diff
@@ -15,7 +15,7 @@ body {
 }
 
 .md-icon-button {
-    margin-right: 10px;
+  	margin-right: 10px;
```
Mctalian: Is that a tab? Please use 4 spaces instead. I'm surprised jscs didn't catch that one.

src/assets/app.css
```diff
@@ -51,13 +51,57 @@ body {
...
+    margin:2px 2px 2px 0px;
+    display:inline-block;
+}
+
+.custom-chip{
+  	margin-bottom: 3px;
+  	padding: 3px 10px;
+  	background-color: #e0e0e0;
+  	color: #424242;
+  	border-radius: 10px;
```
Mctalian: Same with all of these lines, looks like somehow there's a mix of spaces and a tab

src/assets/app.css
```diff
@@ -69,19 +113,20 @@ body {
...
 }
 
 .accent-text {
-    color: #650100;
+  	color: #650100;
 }
 
 .idea-likes-click {
-    cursor: pointer;
+  	cursor: pointer;
```
Mctalian: Check all of the ones above this line too. Maybe your editor is replacing these things? If you're using Atom check to make sure you're using Soft Tabs

alanlgirard: I have soft tabs enabled.  All of the elements only had 2 spaces before.  I went in and changed them all to 4 spaces so it is standardized throughout all of the files.

davethenipper: Should we do any sort of feedback on the colors?  Maybe this is something
that should be demo'd to our group, or even to the Innovation Challenge
members.

Mctalian: Ah, ok. And I'm a derp. jscs wouldn't work on a CSS file :facepunch: 

I think it's ok. Some of them look like hard tabs, but it could be GitHub's fault.

It looks good on my end. Approved.

alanlgirard: Some of those were tabs.  Forgot that I changed settings part way through for something else.  Should be all fixed now

src/assets/app.css
Mctalian: Like here, if you look at lines 54 through 57 and lines 60 through 64, there's a difference in spacing.

The first goup looks like 4 spaces, the second group looks like 2 spaces and a \t

src/assets/app.css
alanlgirard: @davethenipper Yes I am open to feedback on the colors.  I just picked random colors off of the materials colors website.

davethenipper:
Should we do any sort of feedback on the colors?  This might be something we should demo to the Innovation Challenge group.

I need to play around with the colors some to see how they flow with the rest of the site.

alanlgirard:
@davethenipper Yes I am open to feedback on the colors. I just picked random colors off of the materials colors website.

src/assets/app.css
davethenipper: for some reason I'm not seeing this show up in Github... but I'll put this
as an agenda item for the next meeting

Mctalian:
The colors remind me of Skittles. :rainbow: 

Also I looked at the Team Stuff as well (I forgot) and it looks like it works pretty well.

I know I already said "Approved" but when there are no teammates, should there by another message similar to the "You're the only Team mate" message?

Mctalian:
Oh, also if you add someone other than yourself to the team and remove yourself it still tells you that you're the only one on the team even though you're not.

alanlgirard:
Here is a pic of the colors for people that don't want to have to run it.
back colors

YashdalfTheGray:
The colors look good. As the author, should you be able to remove yourself from your own team? Or are we talking about someone other than the author?

Mctalian:
We've discussed what if the author comes up with the idea but they don't
really have the time to run with it? That's where the owner backing tag
came in. Ideally that could be passed to someone else on the team. I'm not
sure if that's the same thought that everyone else on the team has, though.
@davethenipper can we perhaps bring this up briefly in the standup after
reports?

bcbrennecke:
Is that the card I'm working on?  The owner suggesting what type of backing they need

davethenipper:
Sure, I'll add it to the agenda.  And that is not the card you are working on Braun, so no worries.

Also, approved.

alanlgirard:
Added a function to check if a user is a member of the team and cleaned up the text that was being displayed to properly reflect the state.  Everything checks out except for the instance where you navigate to the teams tab while logged out and then log in.  The info doesn't update until a navigation to another tab and back.  I will log an issue so this pull request can get through.

@Mctalian @davethenipper can you re-approve?

YashdalfTheGray:
Is there some place that you can call the refresh idea button? I forget if
there is one or not but I think @Mctalian made one at some point.

Yash Kulshrestha
yash.kulshrestha@gmail.com
(614)563-1163

davethenipper:
Approved!

src/ideas/ideasView/ideasView.js
```diff
@@ -285,6 +295,15 @@ angular.module('flintAndSteel')
                 return false;
             };
 
+            ctrl.isUserMemberOfTeam = function() {
+                ctrl.isInTeam = false;
+                for (var i = 0; i < $scope.idea.team.length; i++) {
+                    if (loginSvc.isUserLoggedIn() && loginSvc.getProperty('_id') === $scope.idea.team[i].memberId) {
+                        ctrl.isInTeam = true;
```
Mctalian: I recommend a `break;` after this line.

Mctalian: On second thought, let's restructure this a bit. Instead of setting a ctrl.isInTeam flag, we can `return true;` in the loop and `return false;` outside of the loop.

Then you can remove this call from `md-on-select` in the HTML, and replace your `ideasView.isInTeam` with this call instead. I tested it locally it works like a charm.

Mctalian: This apparently causes a lot of console errors to appear, though. We'll have to look into this later.

src/ideas/ideasView/ideasView.tpl.html
```diff
@@ -142,54 +145,88 @@
...
-                                            </span>
+                                            {{back.author.name}}
                                         </p>
-                                    </div>
-                                </md-list-item>
-                                <md-list-item class="md-2-line" ng-show="idea.team.length === 0"> <!-- auto add author: needed logic for ||| if there is only one person on the team and it's the author-->
-                                    <div class="md-list-item-text">
-                                        No Team has been formed yet :)
+                                        <span class="custom-chip-wrapper" ng-repeat="type in back.types">
+                                            <small class="custom-chip {{type.name}}">{{type.name}}</small>
```
Mctalian: For the class, the backing types do have `_lowername` properties that currently have no use.

If you got to IdeaSvc.js line ~100, replace the existing type._lowername with this:

`type._lowername = type.name.toLowerCase().replace(/[ ]/g, '-').replace('?', '');`

That takes the name, makes it lowercase, uses regular expressions to find all of the spaces and convert them to dashes, then removes any question marks. Giving you perfect CSS class names, if you want them. Works fine the way it is now.

Mctalian:
At the very least, add the `break;` in the line comments above. Everything else is up to you.

Approved!

Mctalian:
In other words... /play pushit

src/ideas/ideasView/ideasView.tpl.html
alanlgirard: That method is a lot cleaner and works like a charm!  I always like having perfect css classes.  My first method was kind of a hack way of getting it to work.

Mctalian:
Class changes look good and all of the scenarios I tried worked, but we DO get console errors unfortunately. At least on my end. We will look at this tonight and merge later.

Mctalian:
Approved by team.

PR #151, opened 2015-11-18
HTTPS for Open Sans, Ordering Inject index.html, Specifying commits

Mctalian:
Production was blocking the request for the Open Sans font because it was over HTTP not HTTPS. Made that small change.

Also, the reordering of the damn index.html injections was pissing me off, so here's hoping that's finally fixed.

One last thing, I specified commits for projects in the package.json that didn't have Releases or Tags. To make sure they don't update their package and break OUR stuff.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -14,7 +14,8 @@ var gulp = require('gulp'),
     karma = require('karma').server,
     exec = require('child_process').exec,
     mkdirs = require('mkdirs'),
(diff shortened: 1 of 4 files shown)


Mctalian:
@davethenipper @alanlgirard

davethenipper:
Approved!

alanlgirard:
Approve!

PR #152, opened 2015-11-19
Identicons on Toolbar were different from on Account Page

Mctalian:
Also changed the sizes a bit to make them more centered to their white background.

Cleaned up a console error when refreshing on the Account Page.

Fixed a weird JSCS error where it complained about a space at the end of a line.


davethenipper:
Approved

alanlgirard:
Approved

PR #156, opened 2015-11-20
Edit Back

Daviddt17:
This PR adds the ability to edit your back if it already exists.


Mctalian:
Not sure how we're going to address this, but if you edit the back, you can still put in backing types that you've previously input... this may be some sort of 3rd party issue.

image

Also as you're typing the backing types, they appear in the actual back above it. Not a huge deal just a bit confusing especially because there's no cancel button of any kind. This might be passable, we'll see what everyone else thinks:

edit_backs

Thank you for writing the tests, though! Nice work!

Daviddt17:
Yeah I noticed how the types appeared real time. I left it in because I wasn't sure what our desired behavior was in that case. I can easily have it not do that. Also, I think we can handle the duplicate type problem in the controller. Should I try and make the changes?

Mctalian:
I would say "Yes" to fixing the duplication.

And my opinion is also to remove the real-time appending of types to the actual back. Other reviewers, feel free to disagree.

YashdalfTheGray:
I agree with @Mctalian on both accounts. Fix duplication and also remove the real time updates. It looks tacky. 

Also want to commend you on writing tests.

Daviddt17:
So I've fixed the back types updating real time problem. Turns out "$scope.selectedTypes" was referencing "back.types" instead of just copying it.

We may have a problem with the duplicate types. I think there is a problem with md-chips as Bob mentioned earlier. When you preload the chips model, it does not add the preloaded items to the list of things to check for duplicates. However, it does catch duplicates of newly added chips. I tried hacking up a function to run on the adding of a chip, but "md-on-add" is not supported yet in Angular Material 0.11.2.

Any comments on how this should be handled?

Mctalian:
At what version of Angular Material is md-on-add supported?

I have a different idea in the mean-time but I was curious.

Let me test something really quick and get back to you.

Daviddt17:
Looks like 1.0

Mctalian:
Alright so my solution was going to be if Experience is in `$scope.selectedTypes`, don't show it in the autocomplete list, but it doesn't appear to do what I'm expecting.

This makes me think about revisiting changing the Auto-complete to a dialog box with checkboxes or switches for each backing type. We could solve a lot of problems with that, though it could bring along some problems of its own as well.

I'm thinking log an issue and we'll have to address it in the near future. Unless someone can come up with a way to fix this now.

src/ideas/ideasView/ideasView.tpl.html
```diff
@@ -3,7 +3,7 @@
     <md-card-content>
         <div ng-hide="ideasView.enableEdit">
             <h1>{{idea.title}}</h1>
-            <h3>by <email-link author-obj="idea.author"></email-link></h3>
+            <h3>by {{idea.author.name}}</h3>
```
Mctalian: Wait, this is un-doing Matt's recent delivery

src/ideas/ideasView/ideasView.tpl.html
```diff
@@ -78,7 +78,7 @@
                             </div>
                             <div class="md-list-item-text" ng-class="{'deleted': comment.deleted === true}">
                                 <h3>{{comment.text}}</h3>
-                                <p><email-link author-obj="comment.author"></email-link><small class="extra-space-left" ng-bind="momentizeTime(comment.time)"></small></p>
+                                <p>{{comment.author.name}} <small class="extra-space-left" ng-bind="momentizeTime(comment.time)"></small></p>
```
Mctalian: Same here

src/ideas/ideasView/ideasView.tpl.html
```diff
@@ -96,15 +96,21 @@
...
                                     <span class="extra-padding-left-sm">
-                                        <small ng-bind="momentizeTime(back.time)"></small>
-                                    </span>
+                                        <small ng-show="!hasBackBeenEdited(back)" ng-bind="momentizeTime(back.time)"></small>
+                                        <small ng-show="hasBackBeenEdited(back)" ng-bind="momentizeModifiedTime(back.timeModified)"></small>
+                                    </span>                                    
                                 </h3>
                                 <p>
-                                    <email-link author-obj="back.author"></email-link>
+                                    {{back.author.name}}
```
Mctalian: Here as well

src/ideas/ideasView/ideasView.tpl.html
```diff
@@ -135,13 +148,18 @@
                                         <small >{{$chip.name}}</small>
                                     </span>
                                 </md-chip-template>
-                            </md-chips>
+                            </md-chips>
```
Mctalian: Extra whitespace

src/ideas/ideasView/ideasView.tpl.html
```diff
@@ -167,10 +185,7 @@
                                     <div class="md-list-item-text">
                                         <p>
                                             <md-switch ng-model="back.isInTeam" class="md-warn"> <!-- Be careful when editing backs because of this -->
-                                                <email-link author-obj = "back.author"></email-link>
-                                                <span class="extra-space-left">
-                                                    <small class="custom-chip" ng-repeat="type in back.types">{{type.name}}</small>
-                                                </span>
+                                                {{back.author.name}}
```
Mctalian: Un-doing Matt's delivery here as well

src/ideas/ideasView/ideasView.tpl.html
```diff
@@ -204,7 +219,7 @@
                                 <md-list-item class="md-3-line" ng-repeat="back in idea.backs" ng-show="back.isInTeam">
                                     <div class="md-list-item-text">
                                         <p>
-                                            <email-link author-obj = "back.author"></email-link>
+                                            {{back.author.name}}
```
Mctalian: One more

Daviddt17:
I added the issue for potential back type duplicates. Other than that, should we go ahead and merge?

Mctalian:
I approve!  Thanks for making those changes!

Daviddt17:
One more approval needed!

YashdalfTheGray:
Approved!

PR #157, opened 2015-11-20
Adding updates to idea

monotkate:
Disclaimer: I made poor life decisions and stayed up coding too late before an absurdly early flight.  2.5hrs of sleep logic!  Pardon any confusion, just ask.

Added a Status tab that allows anyone that is either a team member or the owner to make status updates.  Went through all the backend stuff to get status into the db.  It reads in very similar to comments.  Team members can delete their own status updates if need be, Idea Owner can delete anyone's.  At some point I want to add the ability to edit those status updates, but I need more sleep first.  I don't think I broke anything?

Status updates are sorted by time - most recent at the top, oldest at the bottom. I pushed the add new form to the top for both status updates and comments.  Both of these should be enabled to confirm on enter now.  I put the form in a card just to make it pop out a bit more since the faded at the top made it kind of hard to catch, but that can easily be changed at a later date.  Oh and @alanlgirard  fixed the time for the comments and so I slipped that in there too (easy fix).  

I think that's it?  I'm going to need sanity checks because.... yeah I need a nap.  I wanted to get this into a pull request though before too many changes went through because of the work I put into the backend stuff to work (which btw, @Mctalian you did a great job of making it relatively easy to figure out, but when you're plowing your way through it trying to figure out what's going on it's a BEAST.  References go all over the place.  Now I know how it works.  Sorta).  Turns out adding the new collection is helpful.   

Yeah.  So that's it I think.

Time for a nap.


Mctalian:
"Status" is that like "updates" on line 30 of the database-schema.json?

I already had all the ground-work set for that :cry: I'm sorry!

Hopefully the actual model stuff and the promises didn't eat up a LOT of that time, because the rest of this looks pretty solid at a glance. Should be able to plug in "updates" into most of the places where you have "statuses". We would just have to add "authorId" to "updates" in the database-schema.json and it should just work. Unless we decide as a team that we want "statuses" or "updates" to live in their own collection. Would like @YashdalfTheGray 's insight on this.

The reason we needed a separate "comments" collection vs. a backs (and now "updates" or "statuses" as well) is because "commments" in the future *might* be nested for replying and such. I think updates should be like backs and be very "flat" so we could keep them tucked under the idea itself instead of referencing another collection.

I just fired it up and it looks great! Love the submitting comments and statuses on "Enter". I actually think Status (or Updates) belongs in the left-most tab. I also really like the form at the top for comments and status. That's a great addition. I'm on the fence as far as the status sorting... Do statuses build on each other? Do we only care about the most recent? I'm now thinking the way you did it is correct.

YashdalfTheGray:
I think updates should live under idea and should be just one level and flat. We can capture the authorId if we want but I think it's understood that the team running the idea is making the updates so we can have a static author field that just says `{{idea.title}} Team` or something like that. I think we should call them updates and they should live in the left-most tab.

database-schema.json
```diff
@@ -50,6 +50,10 @@
         "team": [
             {"memberId": "idea.authorId"},
             "set of objects with user._id"
+        ],
+        "statuses": [
+            {"statusId": "status._id"},
+            "set of objects with status._id"
```
YashdalfTheGray: Updates makes a lot more sense in my mind, the status word is confusing.

database-schema.json
```diff
@@ -68,5 +72,13 @@
...
-    ]
+    ],
+    "status": {
+        "_id": "mongo_generated",
+        "parentId": "one idea._id || one status._id",
+        "text": "",
+        "authorId": "one user._id",
+        "timeCreated": "",
+        "timeModified": ""
+    },
```
YashdalfTheGray: This stuff should live under an idea and will not support nesting.

server/db.js
```diff
@@ -26,17 +26,24 @@ module.exports = function(dbName) {
                             console.log(errEvents);
                         }
                         else {
-                            db.createCollection('comments', function(errComments) {
-                                if (errComments) {
-                                    console.log(errComments);
+                            db.createCollection('statuses', function(errStatuses) {
+                                if (errStatuses) {
+                                    console.log(errStatuses);
```
YashdalfTheGray: We shouldn't need a collection for updates, they are logical children to an idea whereas a comment could have another comment or an idea as a parent.

monotkate:
Goddammit. Oh well learned a lot. Sorry!! Yash I figured it would be nice to have the author listed because if various people are making status updates it might be a good idea to track who is adding what.

server/ideas/ideaModel.js
```diff
@@ -23,6 +23,7 @@ function Idea(title, description, authorId, eventId, tags, rolesreq) {
         types: [{name: "Owner", _lowername: "owner"}]
     }];
     this.team = [{memberId: this.authorId}];
+    this.statuses = [];
```
YashdalfTheGray: Rename to updates,

server/routes.js
```diff
@@ -129,6 +130,33 @@ module.exports = function(app) {
             }
         );
     });
+    app.post('/status', function(req, res) {
```
YashdalfTheGray: This will probably change since updates will wrap up under idea now.

server/routes.js
```diff
@@ -197,6 +225,28 @@ module.exports = function(app) {
             }
         });
     });
+    app.post('/deleteStatus', function(req, res) {
```
YashdalfTheGray: Also this.

monotkate:
I feel like generally you do care only about the most recent status

Mctalian:
I agree about authorId and most recent updates, @monotkate

I think having the author's name under the status update:
A) Makes that member accountable for what they post in updates.
B) Makes being on the team actually mean something.
C) Provides individual visibility to that person in addition to the idea visibility.

Mctalian:
BUT... get some sleep, tackle this later.

YashdalfTheGray:
No, we should care about all the updates and they should still be sorted in reverse chronological order. We can capture the author's name then. You've convinced me. I'm still for renaming status to updates. To me, the status of an idea is just a step but an idea could be at the same step for a while and still generate updates. 

Also don't forget to write unit tests for this functionality.

monotkate:
That makes sense. I like updates. 

Uh oh. I'll figure out the unit tests at some point. I might have time Sunday.

monotkate:
Just commenting to say my best friends mom died (I grew up with her) and I'm heading home early for thanksgiving to grieve with the family. I'll be working from home next week so this will either mean I'll get this done quickly or won't have time to work on it either way. Just a heads up for if I disappear a few days with no updates.

monotkate:
Ok, more plane coding (this is becoming a habit).  @YashdalfTheGray  this should make you happy, it's now renamed back to Updates, flattened back to attach to the idea, and in the first tab.  @Mctalian check out that I didn't miss anything in the switch?  I had to write a routine for deleting, not sure if it's the most efficient, but seems to work.

If you clear performance I'll work on user tests.  Wanted to get functionality down first.

monotkate:
Oh, another thing is that when Alan and I were talking about it yesterday he suggested that the status updates are only visible to team members.  For now I have them as you have to be logged in, but I can easily change that

Mctalian:
I like that anyone can see idea updates. Not even sure we'd need them to
log in. It's letting everyone know how far along the project is and what
blockers they are facing which could enable more people to help them
outside the team. Kinda like how Kickstarter has their updates.

monotkate:
Easy fix. I'll change it. My free 30 mins of Internet are up so I'll post later

Mctalian:
I will check this out locally in a bit.

YashdalfTheGray:
Yeah, I think idea updates should be visible publicly. Also, my condolences, take some time off and go celebrate her life. You've been working too hard anyway :)

LongArmMcGee:
+1

monotkate:
This should have fixed the view issue to not logged in people, and added tests.  And apparently now I have a merge conflict.  noooooooooooo sad day.

Mctalian:
Looks good though! I like it a lot! And yeah, Matt merged in his mailto links earlier today. Hopefully it's not a huge merge conflict! Nice work!

monotkate:
Alright, I think this should be go to go now.  @YashdalfTheGray @Mctalian give me a check?

database-schema.json
Mctalian: This is no longer used, correct? We're using "updates" in the idea collection now?

database-schema.json
Mctalian: Also no longer used?

src/ideas/ideasView/ideasView.js
```diff
@@ -323,6 +361,24 @@ angular.module('flintAndSteel')
                     });
                 }
             };
+
+            ctrl.isUserAuthorOfStatus = function(statusIndex) {
```
Mctalian: Should this be `isUserAuthorOfUpdate` or does this need to get removed? I can't find `isUserAuthorOfUpdate` and it seems to be called in the HTML.

src/ideas/ideasView/ideasView.js
```diff
@@ -323,6 +361,24 @@ angular.module('flintAndSteel')
...
+                if (loginSvc.isUserLoggedIn() && loginSvc.getProperty('_id') === $scope.idea.statuses[statusIndex].authorId) {
+                    return true;
+                }
+                return false;
+            };
+
+            ctrl.deleteStatus = function(statusIndex) {
```
Mctalian: I think this can be removed since we have `deleteUpdate` now.

src/ideas/ideasView/ideasView.js
```diff
@@ -134,15 +135,52 @@ angular.module('flintAndSteel')
...
                     $scope.selectedTypes = [];
                     $scope.selectedType = undefined;
                     ctrl.newComment = '';
                     ctrl.newBack = '';
+                    ctrl.newUpdate = '';
                     ctrl.refreshIdea();
                 }
             };
 
+            $scope.deleteUpdate = function deleteUpdate(index) {
```
Mctalian: Should this have a check `if (isUserAuthorOfUpdate()) {...}`?

src/ideas/ideasView/ideasView.js
```diff
@@ -134,15 +135,52 @@ angular.module('flintAndSteel')
...
+                var numToPop = index;
+                var temp = [];
+                for (var i = 0; i < numToPop; i++) {
+                    temp.push($scope.idea.updates.pop());
```
Mctalian: I think we just want to use [`.splice()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice) here. It would save you a lot of this array manipulation. You can just use something like `$scope.idea.splice(index, 1);` and take out lines 163 thru 174 I think.

src/ideas/ideasView/ideasView.tpl.html
```diff
@@ -67,8 +67,58 @@
...
+                    <md-list>
+                        <md-list-item class="md-2-line" ng-repeat="update in idea.updates | orderBy: '-time'" ng-show="idea.updates" ng-mouseover="hoverUpdate = true" ng-mouseout="hoverUpdate = false">
+                            <div ng-show="(ideasView.isUserAuthorOfUpdate($index) || ideasView.isUserAuthor()) && update.deleted !== true && hoverUpdate === true">
+                                <md-button class="md-icon-button" aria-label="Delete Update" ng-click= "deleteUpdate($index)">
+                                    <i class="material-icons icon-button-adjust">&#xE14C;</i>
+                                </md-button>
+                            </div>
+                            <div class="md-list-item-text" ng-class="{'deleted': update.deleted === true}">
+                                <h3>{{update.text}}</h3>
+                                <p>{{update.author.name}} <small class="extra-space-left" ng-bind="momentizeTime(update.time)"></small></p>
```
Mctalian: We can use Matt's new `email-link` directive for `{{update.author.name}}` here.

```html
<email-link author-obj="update.author"></email-link>
```

src/ideas/ideasView/ideasView.js
monotkate: so I thought about that, but I have it set up so that it's only shown if the user is the author of the update, or if you are the owner of the idea.  Do you still think it needs to check?

Mctalian: I feel like if it's in the html, people can modify it and make calls they shouldn't. (They might be able to do the same with the javascript, so soon we need to change the check for login and stuff to server side), so maybe a revised suggestion:

```javascript
if (isUserAuthorOfUpdate() || isUserAuthor()) {
    ...
}
```

database-schema.json
YashdalfTheGray: Yeah, that's what I thought.

database-schema.json
YashdalfTheGray: Yup.

monotkate:
This should have fixed my issues

Mctalian:
Approved pending removal of the old "statuses" stuff in `database-schema.json`

Mctalian:
Oh no... more conflicts :( this PR is not your friend

monotkate:
PEOPLE NEED TO STOP PUSHING THINGS OMG :-1:

monotkate:
I am tagging everyone and anyone in the hopes that two people will approve this before I need to deal with something else breaking it.

@Mctalian @YashdalfTheGray @LongArmMcGee @alanlgirard @Daviddt17 @davethenipper

Mctalian:
Approve!

bcbrennecke:
Looks good to me, merge it in!

LongArmMcGee:
Glass half full: merge conflicts mean we have multiple developers making
progress

PR #159, opened 2015-11-22
Updating travis to node 4.2.2

YashdalfTheGray:
Officially updating this project to node 4.2.2 which is the current LTS release. The pull request is simple when it comes to this project but what we have to work towards is making sure that everyone is running the right version locally.

So, this is what you do to make sure - 
1. Run `node --version` to make sure it comes back with something close to `4.2.2`. If the version is higher, you should be okay but should bring up any issues that you encounter. If the version is lower, upgrade by installing the LTS version off of the [node](https://nodejs.org/en/) website. 
2. Make sure npm is up to date as well. Run `npm --version` and it should come back with something like `3.4.x`. If it doesn't, run `npm install -g npm` and let npm update itself. After that process is done, run `npm --version` again to verify. 

The automated builds are run against node 4.2.2 and the latest npm version. 

Reviewers requested - @Mctalian @davethenipper 
Action required - @alanlgirard @monotkate @bcbrennecke @LongArmMcGee @Daviddt17


Mctalian:
Just a fairly unimportant comment: I hope they make a v5.x LTS or something like it because I really like the load-bar for npm installs rather than the "/ - \ -" spinny stuff.

Mctalian:
Approved. I'll stay on 5.1.0 until I run into issues. Then I'll report them and revert to 4.2.2 if I have to.

YashdalfTheGray:
That load bar, totally an npm 3.4.x feature, that's why I'm getting everyone to at least update to npm 3.4.x!

Mctalian:
Hmmm, weird. I downgraded to node 4.2.2 yesterday to see if I got node-gyp issues and noticed my load bar went away. I'm now back on 5.1.0 and my load bar is back, but when I read your PR, I checked my npm version and it was like 3.3.x something or other. Maybe each node installer gives you a "just enough" npm by default, not the latest at the time of the release.

monotkate:
I'm updated

LongArmMcGee:
hmmm i'm on node 0.12.2. Not sure how i've been able to run on such an old
version.

YashdalfTheGray:
That was the current version till about 2 months ago when io.js merged into node.js. They then updated their versioning too and they called the merged product node 4. There is also node 5 but what I'm looking for are the performance improvements that came with the io.js merge.

davethenipper:
Approved!  Should the info in the PR be moved somewhere?

YashdalfTheGray:
I think I can put it in the wiki.

alanlgirard:
Approved!  Maybe add the upgrade info to a trello card in the Flint and Steel Internal FAQ board or the wiki works too.

Mctalian:
Agreed. @YashdalfTheGray , can you put something on that trello board?

Mctalian:
Or wherever you see fit. I made a new trello board for internal stuff, it might be a good fit there, or README. Whichever makes more sense.

PR #160, opened 2015-11-22
Load Testing

Mctalian:
This might not be a completely finished thing, but it's a start and I have a decent amount in here I'd like looked over. A lot of commented out code that I want to keep in here for discussion and add or remove based on those discussions.

Some load testing results (on my beast home rig, so we're going to compare just based on this machine, not actual production) (times are in ms, we're creating a bunch of ideas, which I assume is our most expensive operation):

```
// Creating 1000 ideas with up to 100 concurrent connections =========================

With Prints:
stats:  { 
    totalElapsed: 77959.77049300075,
    main: { 
        meter: { 
            mean: 12.829416280310951,
            count: 1000,
            currentRate: 12.829402746257264,
            '1MinuteRate': 9.224287414972048,
            '5MinuteRate': 2.692730345952206,
            '15MinuteRate': 0.969989352632323 
        },
        histogram: { 
            min: 149.99779698252678,
            max: 28861.268723994493,
            sum: 7604117.8173547685,
            variance: 62961310.99354295,
            mean: 7604.117817354769,
            stddev: 7934.816380581402,
            count: 1000,
            median: 3983.7945050150156,
            p75: 12548.525800995529,
            p95: 23856.181180359417,
            p99: 26215.753896793132,
            p999: 28861.062176696483 
        } 
    }
}

Without Prints:
stats:{  
   totalElapsed:10792.234172016382,
   main:{  
      meter:{  
         mean:92.77938061188817,
         count:1000,
         currentRate:17.687303034056136,
         '1MinuteRate':14.890784321948928,
         '5MinuteRate':3.224597295320308,
         '15MinuteRate':1.0893412213689768
      },
      histogram:{  
         min:65.92011699080467,
         max:2777.9475440084934,
         sum:910179.8460613787,
         variance:500779.59703645774,
         mean:910.1798460613787,
         stddev:707.6578248252878,
         count:1000,
         median:707.6502439975739,
         p75:1298.7039889991283,
         p95:2443.9946953669187,
         p99:2645.086303381324,
         p999:2777.9428721984923
      }
   }
}

Without Prints, with 2 CPU Cluster:
stats:{  
   totalElapsed:8658.801813006401,
   main:{  
      meter:{  
         mean:115.70111924033006,
         count:1000,
         currentRate:4.637772498794275,
         '1MinuteRate':12.680955839789343,
         '5MinuteRate':2.621427423891483,
         '15MinuteRate':0.8786681064233611
      },
      histogram:{  
         min:115.51434201002121,
         max:2523.1089510023594,
         sum:699417.140221715,
         variance:212188.20252553656,
         mean:699.4171402217149,
         stddev:460.638906873417,
         count:1000,
         median:602.0971634984016,
         p75:927.1915557458997,
         p95:1641.4189322233194,
         p99:2040.484250568152,
         p999:2523.061696714343
      }
   }
}


Without Prints, with 4 CPU Cluster:
stats:{  
   totalElapsed:5691.634182989597,
   main:{  
      meter:{  
         mean:176.1596379469766,
         count:1000,
         currentRate:30.483220756161554,
         '1MinuteRate':15.655303615578521,
         '5MinuteRate':3.236289341727317,
         '15MinuteRate':1.0847617606411988
      },
      histogram:{  
         min:20.891888976097107,
         max:1831.0611430108547,
         sum:460528.40507140756,
         variance:163204.7018021807,
         mean:460.5284050714076,
         stddev:403.98601683991575,
         count:1000,
         median:316.34382550418377,
         p75:643.6765332445502,
         p95:1319.4370066404335,
         p99:1619.6282011681794,
         p999:1831.0193092748532
      }
   }
}
```

All of this is with NO BROWER CLIENT actually visiting the page. This brings in some issues. Apparently Server-sent events are a bit expensive for one Core. If we have a large amount of traffic at any point, we could see some performance degradation. If we have at least 2 cores, it cuts "posting 300 ideas with a browser open" times by 66%, 4 cores by 90%.

These loads are all VERY large compared to what I think we'll actually see. But we may want to look into clustering our spare PCs using nginx just in case. It would appear that our bottleneck is number of threads/cores.

The other good thing about cluster and workers is that it increases reliability inherently. Even if somehow the server crashes, it's only one core. So the rest of the site will stay up and the core that crashed will be restarted. This way we can diagnose things.

Also this brought up one of my fears, logging to the console. It's a performance hit. We need to look into actual searchable logging because we will inevitably have errors and we need a paper trail to debug them.

Sorry if this is all over the place, it's late and I am tired.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,9 @@ Running the server in development mode disables LDAP login and HTTPS and falls b
 
 In order to run the server in production mode, you need to have access to the company network for LDAP purposes. You also need to create  `server\secrets\ldapAuth.js` which will hold the credentials needed to bind to the LDAP server. Lastly, you will need to have certificates to run HTTPS. Please look up self-signed certificates in order to properly set up the production environment, or obtain certificates from a trusted Certificate Authority.
 
-Any changes to the code while the server is running will result in a server restart. However, this ***DOES NOT*** automatically re-run client tests or JavaScript analysis commands, so be sure to run them yourself before submitting a PR.
+Any changes to the code while the server is running will result in a server restart. However, this **DOES NOT** automatically re-run client tests or JavaScript analysis commands, so be sure to run them yourself before submitting a PR.
+
+`gulp start:test` will run the server in the test environment. This is identical to `dev`, but it uses a hard-coded port number of `7357` and does not generate data.
 
 ### Testing
 
--- a/README.md
+++ b/README.md
@@ -53,6 +53,8 @@ Any changes to the code while the server is running will result in a server rest
 
 To run the client side tests, execute `gulp test:client` from a terminal window. This automatically runs when a `gulp start:dev` or `gulp start:prod` command is executed.
 
+To run the server load tests, execute `gulp start:test` from a terminal window then open an additional terminal window and enter `gulp test:load`. This will submit a number of ideas to the server and provide timing statistics in milliseconds to the terminal window.
+
 ### Linting and style check
 
 Running `gulp jshint` will lint any JavaScript file under the `src` folder excluding `src/lib` and `server` as well as `gulpfile.js`.
--- a/generateData.js
+++ b/generateData.js
@@ -19,83 +19,55 @@ var DB = new mongodb.Db('flintandsteel-dev', new mongodb.Server('localhost', 270
 DB.open(function(err, db) {
     "use strict";
 
-    db.createCollection('users', function(errUsers) {
-        if (errUsers) {
-            console.log(errUsers);
-        }
-        else {
-            var userObjs = [];
-            userObjs.push(UserModel.create("Guybrush", "Threepwood", "Guybrush Threepwood", "test1", "test1@test.com", "Threepy", "Hippy Lumberjack"));
-            userObjs.push(UserModel.create("Rick", "Sanchez", "Rick Sanchez", "test2", "test2@test.com", "Dirty", "Street Pharmacist"));
-            userObjs.push(UserModel.create("Dick", "Dickerson", "Dick Dickerson", "test3", "test3@test.com", "The Fracker", "Money Bags Oil Man"));
-            var insertedUserIds;
-            db.collection('users').insert(userObjs,
-                function(err, results) {
-                    if (err) {
-                        console.log(chalk.bgRed(err));
-                        return;
-                    }
-                    else {
-                        console.log(chalk.bgGreen('Users created in the users collection.'));
-                        insertedUserIds = results.insertedIds;
-                    }
-                }
-            );
-            db.createCollection('events', function(errEvents) {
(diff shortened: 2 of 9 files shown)


YashdalfTheGray:
Shoot, it does seem like nginx would be a really good idea now. I was also thinking about disabling logging in production applications unless there are errors and once someone reports something. running a dev build to see errors happen. We can also try spawning a log writer on another thread to see if that helps. A message sent to another thread through IPC has to be less expensive than trying to render a message in the console. 

I have an idea for a remote log writer that we can explore if you want me to get started on that.

Mctalian:
I'm not sure if we need to get super complex. Winston seems to be a
commonly used tool. Especially if it's async file writes. We should be
writing to files, in my opinion.

YashdalfTheGray:
Yeah, that's what I was thinking haha! Yeah, we should definitely be writing to a file, persistent logging will become a requirement later on so we might as well have it now.

Mctalian:
For now I disabled everything but the error logs so there's not a huge need
but I think this might be MVP if we want to be maintainable. @davethenipper
to make that call though

YashdalfTheGray:
Yeah, I agree and I was looking over that and thought that it was a good idea! @davethenipper, make the call and I can get started on implementing Winston! I can also look into nginx and running concurrent servers for us.

Mctalian:
I made some performance changes for events and it seems to take the need
for concurrency away a lot. So that one is slightly lower priority now
thankfully

YashdalfTheGray:
:) Yay! Saw that too, and I was thinking nginx for post MVP when our codebase stops changing so much and we get a stable build out. Then we can think about server clustering.

Mctalian:
In the meantime core clustering is easy if we want that.

YashdalfTheGray:
True!

davethenipper:
Fine by me!  It sound like something we would need when we launch so we can actually address issues that occur.

Mctalian:
I think this is ready for review:

@YashdalfTheGray, @alanlgirard and anyone else that feels like looking through it.

YashdalfTheGray:
I like this, I read over a lot of this when you initially posted it. I think this is worth while and should give us some really good data!

Approved.

Mctalian:
Once this one goes through I was able to get LDAPS working and I have a PR ready for that

YashdalfTheGray:
Wow, you're just on it like white on rice :100:

alanlgirard:
Looks good!  Approved! Having the ease of load testing will be very helpful in the future.  We might want to keep a log of load testing times somewhere to track performance especially when new features get added in so we have a better idea of what is causing the slow-downs if we run into any.

Mctalian:
That's a good point. I'll run it on the production machine once we merge LDAPS and we'll keep a log.

PR #168, opened 2015-11-25
Identicons added almost everywhere we see an username

Daviddt17:
# What I added:
I've added identicons everywhere with the exception of the edit team toggles. I wasn't sure if identicons were necessary there as they are already included on the normal team view. If it seems like they should be there just let me know and I'll add them. Also, the identicon on the idea author name looks a little funky to me. Let me know what you think. 

One more note, I'm not sure if we should standardize the size of identicons. I've been using 18 for smallish text, 24 for mediumish text, and 30 for biggish text. That might be something to note if we add identicons elsewhere in the future.

Here are some screenshots so everyone can see the identicons in action.

### Author of an idea:
ideaauthor

### Updates:
ideaupdates

### Comment:
ideacomments

### Back:
ideabacks

### Team:
ideateam

### Edit Team (No identicons):
ideateamedit

### Like list:
idealikelist

### Top ideas:
topideas

### Idea browse:
ideabrowse

@YashdalfTheGray @Mctalian @davethenipper @LongArmMcGee @bcbrennecke @alanlgirard @monotkate

### DIFF
--- a/src/ideas/ideaItem/ideaItem.tpl.html
+++ b/src/ideas/ideaItem/ideaItem.tpl.html
@@ -2,7 +2,13 @@
     <md-card-content layout="row">
     <div flex="85" layout="column">
         <div><strong>{{idea.title}}</strong></div>
-        <div class="accent-text">{{idea.author.name}}</div>
+        <div layout="row">
+            <div layout="row" layout-align="center center">
+                <identicon username='idea.author.username' size='18'></identicon>
+            </div>
+            <span>&nbsp;</span>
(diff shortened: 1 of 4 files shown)


Mctalian:
Wow this is great stuff!

As for the author identicon, does removing "by" make it look any better?

Daviddt17:
Yes, much better.

authornoby

Mctalian:
I approve!

YashdalfTheGray:
This is legit, approved.

Daviddt17:
Merged :)

Issue #164, opened 2015-11-23
Add Interaction Concurrency Issue


Mctalian:
If somehow 2 users submitted a back at the exact same moment (or a like, or an update) (maybe even in an even larger window now that we're trying to improve "event" performance), one of those backs might be ignored. Comments are different because they push or pull to the array of comments in an idea rather than overwrite the existing idea object. Since an idea can have multiple users interacting at the same time, we should adjust liking, backing, and updating an idea to follow the same pattern as commenting does. Which means we'll likely need more REST API endpoints to handle all of these interactions.

Actually we can probably make it generic like it used to be.

If anyone wants to pick this up, talk to me first so I can give some guidance.

Mctalian:
David and I were actually able to prove that this is an issue. We both liked an idea at the same-ish time (without the performance increases for events, so the window should be theoretically very small) and one of our likes did not actually get recorded.

Mctalian:
PR merged, this shouldn't be an issue anymore

PR #169, opened 2015-11-25
Concurrency Issues

Mctalian:
Will close issue #164

As of the time of submitting this PR, the following are atomic database operations:
* Adding a comment to an idea (already implemented)
* Removing a comment to an idea (already implemented)
* Liking an idea
* Unliking an idea
* Adding a back to an idea
* Removing a back from an idea (untested since that functionality doesn't exist)

Things that are still needed:
* Editing a back (this might be challenging)
* Editing a comment (not yet implemented, but that should also be atomic for when it does get implemented. Might be outside scope of this PR)
* Editing an update (not yet implemented, but that should also be atomic for when it does get implemented. Might be outside scope of this PR)
* Adding an update to an idea
* Removing an update from an idea
* Anything team related

Why is there a concurrency issue? Anytime we call IdeaSvc.updateIdea in the client code, we are passing the entire Idea object to the server and overwriting it. So if 2 users submit a comment at the same time, only the second comment will show. We need to be updating tiny sections of the idea instead using atomic operations instead.

More to come, but feel free to have a look now:

@YashdalfTheGray @davethenipper @alanlgirard @monotkate @bcbrennecke @LongArmMcGee @Daviddt17

### DIFF
--- a/database-schema.json
+++ b/database-schema.json
@@ -7,10 +7,7 @@
         "username": "",
         "email": "",
         "nickname": "",
-        "title": "",
-        "likedIdeas": [
-            "set of idea._ids"
-        ]
+        "title": ""
     },
     "idea": {
         "_id": "mongo_generated",
--- a/database-schema.json
+++ b/database-schema.json
@@ -7,7 +7,10 @@
         "username": "",
         "email": "",
         "nickname": "",
-        "title": ""
+        "title": "",
+        "likedIdeas": [
+            "set of idea._ids"
+        ]
     },
     "idea": {
         "_id": "mongo_generated",
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -138,7 +138,10 @@ gulp.task('start:dev', ['test:client', 'inject', 'generate:data'], function() {
     nodemon({
         script: 'server/server.js',
         env: { 'NODE_ENV': 'development' },
-        'ignore': ['server/datastore/*']
+        'ignore': [
+            'server/datastore/*',
+            'src/*'
+        ]
     });
 });
 
@@ -148,7 +151,10 @@ gulp.task('start:test', ['test:client', 'inject'], function() {
     nodemon({
         script: 'server/server.js',
         env: { 'NODE_ENV': 'test' },
-        'ignore': ['server/datastore/*']
+        'ignore': [
+            'server/datastore/*',
+            'src/*'
+        ]
     });
 });
 
@@ -158,7 +164,10 @@ gulp.task('start:prod', ['test:client', 'inject'], function() {
     nodemon({
         script: 'server/server.js',
         env: { 'NODE_ENV': 'production' },
-        'ignore': ['server/datastore/*']
+        'ignore': [
+            'server/datastore/*',
+            'src/*'
+        ]
     });
 });
 
--- a/server/db.js
+++ b/server/db.js
@@ -175,6 +175,34 @@ module.exports = function(dbName, cb) {
         );
     };
 
+    module.updateOneArrayElement = function updateArrayElement(collection, id, property, searchProperty, searchValue, newValue, cb) {
+        var objId = new ObjectId(id);
+
+        var query = {};
+        query._id = objId;
+        query[property + "." + searchProperty] = searchValue; // Find the old value
+
+        var update = {};
+        update.$set = {};
+        // '.$' Update first match of query (should be only match)
+        update.$set[property + ".$"] = newValue;
+
+        db.collection(collection).updateOne(
+            query,
+            update,
+            function(err, results) {
+                if (err) {
+                    console.error(chalk.bgRed(err));
+                    cb(err);
+                }
+                else {
+                    // console.log(chalk.bgGreen('Document with id %s updated in the ' + collection + ' collection.'), id);
+                    cb(null, results);
+                }
+            }
+        );
+    };
+
     module.updateOnePullArray = function updateOnePushArray(collection, id, property, value, cb) {
 
         var objId = new ObjectId(id);
--- a/server/db.js
+++ b/server/db.js
@@ -175,6 +175,32 @@ module.exports = function(dbName, cb) {
         );
     };
 
+    module.updateOnePullArray = function updateOnePushArray(collection, id, property, value, cb) {
+
+        var objId = new ObjectId(id);
+
+        var update = {};
+        update[property] = value;
+
+        var pull = {};
+        pull.$pull = update;
+
+        db.collection(collection).updateOne(
+            { _id: objId },
+            pull,
+            function(err, results) {
+                if (err) {
+                    console.error(chalk.bgRed(err));
+                    cb(err);
+                }
+                else {
+                    // console.log(chalk.bgGreen('Document with id %s updated in the ' + collection + ' collection.'), id);
+                    cb(null, results);
+                }
+            }
+        );
+    };
+
     module.findAndPullArray = function findAndPullArray(collection, property, value, cb) {
         var find = {};
         find[property] = value;
--- a/server/ideas/ideas.js
+++ b/server/ideas/ideas.js
@@ -140,8 +140,7 @@ module.exports = function(db) {
             setTimeout(function() {
                 ideaProto.emit("newHeaders", newestHeaders);
                 isEmittingHeaders = false;
-            }
-            , 500);
+            }, 500);
         }
     };
 
@@ -168,8 +167,7 @@ module.exports = function(db) {
             setTimeout(function() {
                 ideaProto.emit("updateIdea_" + key, idea);
                 isEmittingUpdates = false;
-            }
-            , 500);
+            }, 500);
         }
     };
 
--- a/server/ideas/ideas.js
+++ b/server/ideas/ideas.js
@@ -29,8 +29,10 @@ module.exports = function(db) {
     };
 
     module.like = function(id, userId, cb) {
+        var objId = new ObjectId(userId);
+
         var obj = {
-            userId: userId
+            userId: objId
         };
         db.updateOnePushArray(COLLECTION, id, "likes", obj, cb);
     };
@@ -42,9 +44,7 @@ module.exports = function(db) {
             userId: objId
         };
 
-        db.findAndPullArray(COLLECTION, "likes", obj, function(err, results) {
-            cb(err, results);
-        });
+        db.updateOnePullArray(COLLECTION, id, "likes", obj, cb);
     };
 
     module.update = function(id, property, value, cb) {
--- a/server/ideas/ideas.js
+++ b/server/ideas/ideas.js
@@ -47,6 +47,14 @@ module.exports = function(db) {
         db.updateOnePullArray(COLLECTION, id, "likes", obj, cb);
     };
 
+    module.back = function(id, backObj, cb) {
+        db.updateOnePushArray(COLLECTION, id, "backs", backObj, cb);
+    };
+
+    module.unback = function(id, userId, cb) {
+        db.updateOnePullArray(COLLECTION, id, "backs", backObj, cb);
+    };
+
     module.update = function(id, property, value, cb) {
         var updateObj = {};
         updateObj[property] = value;
--- a/server/ideas/ideas.js
+++ b/server/ideas/ideas.js
@@ -40,28 +40,22 @@ module.exports = function(db) {
         db.updateOnePushArray(COLLECTION, id, interactionType, interactionObject, cb);
     };
 
-    module.unlike = function(id, userId, cb) {
-        var objId = new ObjectId(userId);
+    module.removeInteraction = function(id, interactionType, interactionObject, cb) {
+        if (typeof interactionObject.userId !== "undefined") {
+            interactionObject.userId = new ObjectId(interactionObject.userId);
+        }
 
-        var obj = {
-            userId: objId
-        };
+        if (typeof interactionObject.authorId !== "undefined") {
+            interactionObject.authorId = new ObjectId(interactionObject.authorId);
+        }
 
-        db.updateOnePullArray(COLLECTION, id, "likes", obj, cb);
-    };
-
-    module.unback = function(id, backObj, cb) {
-        db.updateOnePullArray(COLLECTION, id, "backs", backObj, cb);
+        db.updateOnePullArray(COLLECTION, id, interactionType, interactionObject, cb);
     };
 
     module.editBack = function(id, authorId, newBack, cb) {
         var objId = new ObjectId(authorId);
         newBack.authorId = new ObjectId(newBack.authorId);
         db.updateOneArrayElement(COLLECTION, id, "backs", "authorId", objId, newBack, cb);
-    };
-
-    module.deleteUpdate = function(id, updateObj, cb) {
-        db.updateOnePullArray(COLLECTION, id, "updates", updateObj, cb);
     };
 
     module.update = function(id, property, value, cb) {
@@ -203,7 +197,7 @@ module.exports = function(db) {
         if (!isEmittingUpdates) {
             isEmittingUpdates = true;
             setTimeout(function() {
-                ideaProto.emit("updateIdea_" + key, idea);
+                ideaProto.emit("updateIdea_" + key, newestIdea);
                 isEmittingUpdates = false;
             }, 500);
         }
--- a/server/ideas/ideas.js
+++ b/server/ideas/ideas.js
@@ -55,6 +55,12 @@ module.exports = function(db) {
         db.updateOnePullArray(COLLECTION, id, "backs", backObj, cb);
     };
 
+    module.editBack = function(id, authorId, newBack, cb) {
+        var objId = new ObjectId(authorId);
+        newBack.authorId = new ObjectId(newBack.authorId);
+        db.updateOneArrayElement(COLLECTION, id, "backs", "authorId", objId, newBack, cb);
+    };
+
     module.postUpdate = function(id, updateObj, cb) {
         db.updateOnePushArray(COLLECTION, id, "updates", updateObj, cb);
     };
--- a/server/ideas/ideas.js
+++ b/server/ideas/ideas.js
@@ -55,6 +55,14 @@ module.exports = function(db) {
         db.updateOnePullArray(COLLECTION, id, "backs", backObj, cb);
     };
 
+    module.postUpdate = function(id, updateObj, cb) {
+        db.updateOnePushArray(COLLECTION, id, "updates", updateObj, cb);
+    };
+
+    module.deleteUpdate = function(id, updateObj, cb) {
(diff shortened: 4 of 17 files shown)


database-schema.json
```diff
@@ -7,7 +7,10 @@
         "username": "",
         "email": "",
         "nickname": "",
-        "title": ""
+        "title": "",
+        "likedIdeas": [
+            "set of idea._ids"
+        ]
```
Mctalian: Simply updated the schema json to reflect what we were already doing.

gulpfile.js
```diff
@@ -158,7 +164,10 @@ gulp.task('start:prod', ['test:client', 'inject'], function() {
     nodemon({
         script: 'server/server.js',
         env: { 'NODE_ENV': 'production' },
-        'ignore': ['server/datastore/*']
+        'ignore': [
+            'server/datastore/*',
+            'src/*'
+        ]
```
Mctalian: All of these are adding client code to the ignore for restarting the server on changes. Client side code is served up as requested so we don't need to restart the server for changes in any of those files.

server/db.js
```diff
@@ -175,6 +175,32 @@ module.exports = function(dbName, cb) {
...
+                    cb(err);
+                }
+                else {
+                    // console.log(chalk.bgGreen('Document with id %s updated in the ' + collection + ' collection.'), id);
+                    cb(null, results);
+                }
+            }
+        );
+    };
+
```
Mctalian: This makes removing an element from an array within an object (a.k.a. Likes, Backs, Updates) atomic so that if two happen at the same time, they do not interfere with each other.

server/ideas/ideas.js
```diff
@@ -28,6 +28,33 @@ module.exports = function(db) {
...
+    };
+
+    module.back = function(id, backObj, cb) {
+        db.updateOnePushArray(COLLECTION, id, "backs", backObj, cb);
+    };
+
+    module.unback = function(id, backObj, cb) {
+        db.updateOnePullArray(COLLECTION, id, "backs", backObj, cb);
+    };
+
```
Mctalian: New methods for liking, unliking, backing, and unbacking an idea. They use the atomic db operations for pushing to and pulling from arrays.

server/routes.js
```diff
@@ -146,6 +146,84 @@ module.exports = function(app) {
...
+                ideas.get(req.body.id, function(err, idea) {
+                    IdeasInstance.updateIdea(idea);
+                });
+                ideas.fetch(function(err, headers) {
+                    IdeasInstance.newHeaders(headers);
+                });
+                res.sendStatus(200);
+            }
+        });
+    });
```
Mctalian: API end-points that handle requests to like, unlike, back, or unback an idea.

Likes and unlikes are special because we have to update a user object as well since we track "likedIdeas"

server/server.js
```diff
@@ -39,7 +39,7 @@ app.use(express.static(path.join(__dirname + '/../src')));
 app.use(bodyParser.json());
 
 if (process.env.NODE_ENV === 'production') {
-
+    process.env.NODE_TLS_REJECT_UNAUTHORIZED = "0";
```
Mctalian: Needed for LDAPS stuff. It complained on my laptop but not in previous tests. Weird.

server/server.js
```diff
@@ -66,25 +66,21 @@ if (process.env.NODE_ENV === 'production') {
...
-        if (id) {
-            done(null);
+        var users = require('./users/users')(GLOBAL.db);
+
+        if (user) {
+            console.log('deserializeUser:' + user._json.sAMAccountName);
+            users.findForLogin(user, done);
+        }
+        else {
+            done("No user provided!");
```
Mctalian: Serlalizing and Deserializing is used for Production when we try to enable sessions and keeping people logged in if they refresh the page. I have no idea if I did this right and have no way to test it right now.

server/users/users.js
```diff
@@ -81,7 +81,8 @@ module.exports = function(db) {
                 var responseObj = {
                     name: doc.fullName,
                     mail: doc.email,
-                    username: doc.username
+                    username: doc.username,
+                    likedIdeas: doc.likedIdeas
```
Mctalian: We might use likedIdeas in user headers later. Maybe for "user" pages or the account page or something. I thought i needed it for this delivery but I didn't. Alas, it's still here.

server/users/users.js
```diff
@@ -97,5 +98,13 @@ module.exports = function(db) {
...
     };
 
+    module.likeIdea = function likeIdea(id, ideaId, cb) {
+        db.updateOnePushArray(COLLECTION, id, "likedIdeas", ideaId, cb);
+    };
+
+    module.unlikeIdea = function unlikeIdea(id, ideaId, cb) {
+        db.updateOnePullArray(COLLECTION, id, "likedIdeas", ideaId, cb);
+    };
+
```
Mctalian: Adding and removing ideas from the user's likedIdeas.

src/ideas/ideaSvc/ideaSvc.js
```diff
@@ -61,6 +61,58 @@ angular.module('flintAndSteel')
...
+                            {
+                                id: ideaId,
+                                backObj: backObj
+                            }
+                        )
+                        .success(successCb)
+                        .error(errorCb);
+                }
+            };
+
```
Mctalian: Client-side requests to the server for liking, unliking, backing, unbacking (we haven't implemented deleting backs yet, so this is untested).

src/ideas/ideaSvc/ideaSvc.js
```diff
@@ -111,6 +163,10 @@ angular.module('flintAndSteel')
                 deleteComment: this.deleteComment,
                 getUniqueId: this.getUniqueId,
                 updateIdea: this.updateIdea,
+                likeIdea: this.likeIdea,
+                unlikeIdea: this.unlikeIdea,
+                backIdea: this.backIdea,
+                unbackIdea: this.unbackIdea,
```
Mctalian: Exposing the new methods to the public interface. We're returning an object and that object has these methods. So when someone uses `ideaSvc` in code, they can call any of the methods listed here.

src/ideas/ideaSvc/ideaSvc.mock.js
```diff
@@ -23,21 +23,13 @@ angular.module('flintAndSteel')
...
-                    'anaerobedigits',
-                    'chouxthames',
-                    'pizzago'
+                    {userId: 1},
+                    {userId: 2},
+                    {userId: 3},
+                    {userId: 4},
+                    {userId: 5},
+                    {userId: 6},
+                    {userId: 7}
```
Mctalian: Updated a mock to represent the schema a bit better.

src/ideas/ideaSvc/ideaSvc.mock.js
```diff
@@ -151,13 +135,26 @@ angular.module('flintAndSteel')
...
+                    successCb('OK');
+                },
+                backIdea: function backIdea(ideaId, backObj, successCb) {
+                    mockIdea.backs.push(backObj);
+                    successCb('OK');
+                },
+                unbackIdea: function unbackIdea(ideaId, backObj, successCb) {
+                    mockIdea.backs.splice(mockIdea.backs.indexOf(backObj), 1);
+                    successCb('OK');
+                },
```
Mctalian: Added the new methods to the mock.

src/ideas/ideaSvc/ideaSvc.mock.js
```diff
@@ -179,7 +176,7 @@ angular.module('flintAndSteel')
                         { name: 'Test Chip'}
                     ];
                     return types.map(function(type) {
-                        type._lowername = type.name.toLowerCase();
+                        type._lowername = type.name.toLowerCase().replace(/[ ]/g, '-').replace('?', '');
```
Mctalian: This is what we actually do in production, so I update the mock.

src/ideas/ideasView/ideasView.js
```diff
@@ -128,19 +128,12 @@ angular.module('flintAndSteel')
...
                             types: $scope.selectedTypes
                         });
-
-                        ideaSvc.updateIdea($scope.idea._id, type, $scope.idea[type],
-                            function success() { },
-                            function error(data, status) {
-                                console.log(status);
-                            }
-                        );
                     }
```
Mctalian: Not using `updateIdea` anymore. We're calling through to use the atomic db operations and passing the whole back object to get inserted seamlessly into this idea's backs.

src/ideas/ideasView/ideasView.js
```diff
@@ -178,33 +171,52 @@ angular.module('flintAndSteel')
...
+
+            $scope.unbackIdea = function unbackIdea(backObj) {
+                ideaSvc.unbackIdea($scope.idea._id, backObj,
+                    function success() {
+                        ctrl.refreshIdea();
+                    },
+                    function error(data, status) {
+                        console.log(status);
+                    }
+                );
```
Mctalian: Call the new service methods so they can call the new server-side end points.

src/ideas/ideasView/ideasView.js
```diff
@@ -178,33 +171,52 @@ angular.module('flintAndSteel')
...
             };
 
             $scope.isUserLiked = function isUserLiked() {
-                var likedIdeas = loginSvc.getProperty('likedIdeas');
-                //console.log(likedIdeas);
-                return (_.findIndex(likedIdeas, function(item) { return item === $scope.idea._id; }) !== -1);
+                var userId = loginSvc.getProperty('_id');
+                return (_.findIndex($scope.idea.likes, function(like) { return like.userId === userId;}) !== -1);
```
Mctalian: Find out if a user liked an idea in a slightly different way because we're tracking likedIdeas on the server side now, not the client side.

src/ideas/ideasView/ideasView.js
```diff
@@ -411,7 +423,7 @@ angular.module('flintAndSteel')
                     $scope.idea.backs.forEach(function(back) {
                         if (loginSvc.getProperty('_id') === back.authorId) {
                             hasUserBacked = true;
-                        }                        
+                        }
```
Mctalian: My editor is picky and removed some whitespace for me.

src/users/loginSvc/loginSvc.js
```diff
@@ -100,14 +53,10 @@ angular.module('flintAndSteel')
...
                 isUserLoggedIn: this.isUserLoggedIn,
                 logout: this.logout,
                 getProperty: this.getProperty,
-                likeIdea: this.likeIdea,
-                unlikeIdea: this.unlikeIdea,
                 updateAccount: this.updateAccount,
-                checkValidUsername: this.checkValidUsername,
                 getUserById: this.getUserById
             };
         }
```
Mctalian: Removed old, unused code.

src/users/loginSvc/loginSvc.mock.js
```diff
@@ -44,18 +41,9 @@ angular.module('flintAndSteel')
...
-                },
                 updateAccount: function updateAccount() {
                     throw new NotImplementedException('updateAccount');
                 },
-                checkValidUsername: function checkValidUsername() {
-                    throw new NotImplementedException('checkValidUsername');
-                },
                 getUserById: function getUserById() {
                     return mockUserAccount;
                 }
```
Mctalian: Removed old, unused code.

src/users/loginSvc/loginSvc.spec.js
```diff
@@ -202,34 +137,4 @@ describe('loginSvc', function() {
...
-
-            loginSvc.checkValidUsername('dummy', function(data) {
-                expect(data).toBe(false);
-            }, function() { });
-
-            $httpBackend.flush();
-        });
-    });
-
 });
```
Mctalian: Removed old, unused code.

server/routes.js
YashdalfTheGray: With mongo, I don't think we need to track liked ideas, we should be able to query every idea that a user has liked from the server without needing to store it.

server/users/users.js
YashdalfTheGray: This shouldn't be necessary, the more I think about it, the more I realize that tracking likedIdeas under a user is just an added pain.

Mctalian: We were doing it already. Rather than change algorithms in this PR, I just
kept status quo.

YashdalfTheGray:
Alright, so I've been thinking of and kind of working on making our backend totally RESTful ever since the mongo PR went in. Maybe after this, I'll go in and make sure that our API is not only RESTful but also uses CRUD rather than just getting and posting all the time.

That being said...

That's not how an HTTP PATCH should be used. You nailed the atomic part, PATCH is probably the only guaranteed atomic action in all of the HTTP verbs. But PATCH is used to apply a diff to an HTTP endpoint. The endpoint that is listening to a patch is supposed to look for an operation, a path and a value. So to update an object, you would pass something like this as the payload of a PATCH verb
```json
[
    { "operation": "update", "path": "a/b/c", "value": "something" },
    { "operation": "delete", "path": "a/e" },
    { "operation" "create", "path": "a/b/d", "value": { "prop": "value" } },
]
```
and the PATCH will make sure that all of those operations are run on whatever endpoint you call patch on. Granted, PATCH doesn't a specific way of configuring that payload so most people don't pay too much attention to this and even more tend to ignore how it's supposed to work but there is a standard for expressing a PATCH payload in JSON which looks pretty similar to what I wrote above. 

All of the patch endpoints need to be reworked like that. I can help with this if you want, there are other API endpoints that need to be cleaned up too.

Mctalian:
I'll just use post for now and leave the CRUD, REST stuff to you. There's
too much in this PR already

Mctalian:
@YashdalfTheGray are you sure what you're describing above is PATCH? Everything I have read has been similar to this:

From http://restcookbook.com/HTTP%20Methods/patch/ :

When should we use the PATCH HTTP method?

The HTTP methods PATCH can be used to update partial resources. For instance, when you only need to update one field of the resource, PUTting a complete resource representation might be cumbersome and utilizes more bandwidth

PATCH /user/jthijssen HTTP/1.1
<user>
    <firstname>Joshua</firstname>
</user>
Also, the PUT method is idempotent. PUTting the same data multiple times to the same resource, should not result in different resources, while POSTing to the same resource can result in the creation of multiple resources.

- See more at: http://restcookbook.com/HTTP%20Methods/patch/#sthash.9v58lVS8.dpuf

Mctalian:
Update:

* ~~Adding a comment to an idea (already implemented)~~
* ~~Removing a comment to an idea (already implemented)~~
* ~~Liking an idea~~
* ~~Unliking an idea~~
* ~~Adding a back to an idea~~
* ~~Removing a back from an idea (untested since that functionality doesn't exist)~~
* ~~Adding an update to an idea~~
* ~~Removing an update from an idea~~

Things that are still needed:

Editing a back (this might be challenging)
Editing a comment (not yet implemented, but that should also be atomic for when it does get implemented. Might be outside scope of this PR)
Editing an update (not yet implemented, but that should also be atomic for when it does get implemented. Might be outside scope of this PR)
Anything team related

YashdalfTheGray:
The spec for PATCH from IETF say that it's a set of operations that should be performed on an endpoint. There is another [article](http://williamdurand.fr/2014/02/14/please-do-not-patch-like-an-idiot/) that I found that says that the common misconception is using PATCH like PUT but it's supposed to be a set of instructions instead.

bcbrennecke:
Holy heck.  Two words.  Thanksgiving.  Break. @mctalian @yashdalfthegray

YashdalfTheGray:
Fair enough, @Mctalian, wanna talk about this on Monday? We can also take this offline if you want.

Mctalian:
I don't care about PATCH,  I care more about getting PRs through. You
handle CRUD and REST. I'll finish this up in a few hopefully and then I'll
enjoy my thanksgiving break and people can review it at their leisure.

YashdalfTheGray:
Alright, sounds good.

Mctalian:
Update:

* ~~Adding a comment to an idea (already implemented)~~
* ~~Removing a comment to an idea (already implemented)~~
* ~~Liking an idea~~
* ~~Unliking an idea~~
* ~~Adding a back to an idea~~
* ~~Removing a back from an idea (untested since that functionality doesn't exist)~~
* ~~Adding an update to an idea~~
* ~~Removing an update from an idea~~
* ~~Editing a back~~

Things that are still needed:

* Editing a comment (not yet implemented, but that should also be atomic for when it does get implemented. Might be outside scope of this PR)
* Editing an update (not yet implemented, but that should also be atomic for when it does get implemented. Might be outside scope of this PR)
* Anything team related

Mctalian:
I did not refactor the "Editing" things into their own method. I don't see editing of comments or updates in the MVP, so I think this is good enough for now. Something to keep in mind for the future.

server/routes.js
Mctalian: No longer tracking likedIdeas as a result of my most recent commit.

server/routes.js
```diff
@@ -198,16 +246,6 @@ module.exports = function(app) {
...
-    app.post('/updateaccount', function(req, res) {
-        users.update(req.body._id, "likedIdeas", req.body.likedIdeas, function(err, results) {
-            if (err || results.value === null) {
-                console.error(chalk.bgRed(err));
-            }
-            else {
-                res.sendStatus(200);
-            }
-        });
-    });
```
Mctalian: This was only used for "likedIdeas" so I removed it and all other references to "likedIdeas"

src/ideas/ideasView/ideasView.js
Mctalian: We're no longer tracking likedIdeas at all so we needed to do this in a different way anyway.

Mctalian:
This is reviewable now. Sorry I didn't make that clear. If you have any questions about a line or a commit, I can add some commentary. Need 2 reviewers:

@YashdalfTheGray @davethenipper @alanlgirard @monotkate @bcbrennecke @LongArmMcGee @Daviddt17

gulpfile.js
```diff
@@ -148,7 +151,10 @@ gulp.task('start:test', ['test:client', 'inject'], function() {
     nodemon({
         script: 'server/server.js',
         env: { 'NODE_ENV': 'test' },
-        'ignore': ['server/datastore/*']
+        'ignore': [
+            'server/datastore/*',
+            'src/*'
+        ]
```
YashdalfTheGray: Can you add something to the readme saying that when you update client code to do a hard refresh?

server/routes.js
YashdalfTheGray: :+1:

YashdalfTheGray:
Approved! Big fan of the login simplification :)

Mctalian:
Updated the Readme as requested and rebased to latest.

monotkate:
What I see looks good.  I like it.  Good changes.  Approved

PR #171, opened 2015-11-27
Fixed order of updates - doesn't always reroute to home on logout

monotkate:
So this is pretty minor. 
(1) I realized that I'd goofed the order of updates on one of my fixes so I fixed that, the most recent update now presents at the top.

(2) I was getting really irritated during testing that it always re-routed to home on logout.  What I added was refreshing the page if there was potentially edits going on during a logout which should wipe any data left unfinished by reloading the controller.  It will reload ideaView and reroute addidea to home.  Any other page it will leave untouched since is far as I'm aware there should be no accessible data that would need to be cleared.  If people don't like this idea or I'm missing something, just let me know.  

### DIFF
--- a/src/users/accountView/accountView.js
+++ b/src/users/accountView/accountView.js
@@ -28,13 +28,7 @@ angular.module('flintAndSteel')
 
                 toastSvc.show(accountName + ' has been logged out!', { duration: 5000 });
 
-                // Reload if editing data to clear partial content
-                if ($state.includes('idea')) {
-                    $state.reload('idea');
-                }
-                else if ($state.includes('addidea')) {
-                    $state.go('home');
-                }
+                $state.go('home');
             };
         }
     ]
--- a/src/users/accountView/accountView.js
+++ b/src/users/accountView/accountView.js
@@ -28,7 +28,13 @@ angular.module('flintAndSteel')
 
                 toastSvc.show(accountName + ' has been logged out!', { duration: 5000 });
 
-                $state.go('home');
+                // Reload if editing data to clear partial content
+                if ($state.includes('idea')) {
+                    $state.reload('idea');
+                }
+                else if ($state.includes('addidea')) {
+                    $state.go('home');
+                }
             };
         }
     ]
--- a/src/ideas/ideasView/ideasView.tpl.html
+++ b/src/ideas/ideasView/ideasView.tpl.html
@@ -89,7 +89,7 @@
                         </md-card-content>
                     </md-card>
                     <md-list>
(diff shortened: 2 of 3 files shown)


Mctalian:
As far as the update order, I think we were discussing it during your PR
that it'd be better to have the Updates in chronological order because
updates build off each other, much like the KickStarter updates. Making
each update important, not just the most recent.

monotkate:
No, Yash ended up saying that we should care about all the updates, but they should be in reverse chronological order.  I just checked to make sure.

Mctalian:
Hmm memory didn't serve me correct. Either way, David and I found an issue
with deleting updates when there were a few updates, the original author
couldn't delete their own, we might have changed the order to get our
bearings to fix it then forgot to switch it back. So just make sure
everything functions as expected and we'll be good!

monotkate:
Alright, I'll check it probably tomorrow.  Switched comps to work on my other project for now.  I altered things to delete in the right order when they displayed in reverse order, but I think one of the merge conflicts made me change the time tag and accidentally get the order wrong.  I think once the order's fixed, the delete should work correctly

src/users/accountView/accountView.js
```diff
@@ -28,7 +28,13 @@ angular.module('flintAndSteel')
...
                 toastSvc.show(accountName + ' has been logged out!', { duration: 5000 });
 
-                $state.go('home');
+                // Reload if editing data to clear partial content
+                if ($state.includes('idea')) {
+                    $state.reload('idea');
+                }
+                else if ($state.includes('addidea')) {
+                    $state.go('home');
+                }
```
Mctalian: If we're logging out from the accountView, I think we're already in the "account" state. So I think we actually do want to go "home" from here.

Mctalian:
I changed how updates are deleted on the client-side. It no longer relies on the index of the update, but rather passes in the update object itself. You can then order updates however you want without changing your deleting algorithm.

So don't worry about it anymore, we just need to make sure either my PR goes in first, or mine is shortly after yours or I think the build will break.

monotkate:
Gotcha. Mine should be a quick check. Not a lot of code added. I'm not getting on a computer tonight though and idk about tomorrow.

Mctalian:
Yeah this looks good. Take a look at my comment on accountView when you get a chance, other than that, I approve.

src/users/accountView/accountView.js
monotkate: Good call.  Didn't think that one through just did a copy/paste.  Posting an updated commit for that in a sec

monotkate:
Can I get another approve on this? @YashdalfTheGray @alanlgirard

YashdalfTheGray:
Approved!

PR #190, opened 2015-12-04
email team feature

LongArmMcGee:
Add an email team button.

### DIFF
--- a/src/ideas/ideasView/ideasView.js
+++ b/src/ideas/ideasView/ideasView.js
@@ -534,10 +534,11 @@ angular.module('flintAndSteel')
 
             ctrl.parseTeamEmail = function parseTeamEmail() {
                 var emailString = "mailto:";
-                $scope.idea.team.forEach(function(teamElement) {
-                    // emailString + memberId.getProperty('email') + ';';
-                    emailString += teamElement.member.mail + ';';
-                });
+                if (angular.isDefined($scope.idea.team)) {
+                    $scope.idea.team.forEach(function(teamElement) {
+                        emailString += teamElement.member.mail + ';';
+                    });
+                };
                 return emailString;
             };
 
--- a/src/ideas/ideasView/ideasView.js
+++ b/src/ideas/ideasView/ideasView.js
@@ -397,18 +397,15 @@ angular.module('flintAndSteel')
                     $scope.selectedTypes = [];
                 }
             };
-<<<<<<< HEAD
 
             ctrl.parseTeamEmail = function parseTeamEmail() {
                 var emailString = "mailto:";
-                $scope.idea.team.forEach(function(memberId) {
+                $scope.idea.team.forEach(function(teamElement) {
                     // emailString + memberId.getProperty('email') + ';';
-                    emailString + memberId.member.email + ';';
+                    emailString += teamElement.member.mail + ';';
                 });
                 return emailString;
             };
-=======
->>>>>>> refs/remotes/YashdalfTheGray/master
 
             $scope.hasUserBacked = function() {
                 var hasUserBacked = false;
--- a/src/ideas/ideasView/ideasView.js
+++ b/src/ideas/ideasView/ideasView.js
@@ -538,7 +538,7 @@ angular.module('flintAndSteel')
                     $scope.idea.team.forEach(function(teamElement) {
                         emailString += teamElement.member.mail + ';';
                     });
-                };
+                }
                 return emailString;
             };
 
--- a/src/ideas/ideasView/ideasView.js
+++ b/src/ideas/ideasView/ideasView.js
@@ -409,7 +409,15 @@ angular.module('flintAndSteel')
                     $scope.selectedTypes = [];
                     ctrl.refreshIdea();
                 }
-            };            
+            };
+
+            ctrl.parseTeamEmail = function parseTeamEmail() {
+                emailString = "mailto:";
+                $scope.idea.team.memberId.forEach(function(memberId) {
+                    emailString + memberId.getProperty('email') + ';';
+                });
+                return emailString;
+            };
 
             $scope.hasUserBacked = function() {
                 var hasUserBacked = false;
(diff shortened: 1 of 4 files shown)


LongArmMcGee:
Approve me, plz

YashdalfTheGray:
Are you actually using `src/assets/ic_email_black_24px.svg`?

Mctalian:
You didn't end up putting that email team button in the team tab in addition to the more subtle yet accessible alternative near the idea title.

I'll also take a look at how it looks tomorrow, but I noticed that part from the code. I agree with Yash, I don't think you're using that svg.

LongArmMcGee:
good eye, nerds. Icon removed.

LongArmMcGee:
I noticed an issue.. my md-icon is no longer clickable. Using md-button instead puts it way high and further left though... not sure how to correct that.

src/ideas/ideasView/ideasView.tpl.html
```diff
@@ -2,7 +2,16 @@
     <img ng-if="ideaHasImage()" ng-src="{{idea.image}}" class="md-card-image">
     <md-card-content>
         <div ng-hide="ideasView.enableEdit">
-            <h1>{{idea.title}}</h1>
+            <div>
+            <h1>{{idea.title}}
+                <span>
+                    <md-icon aria-label="email group" class="md-icon-button small-padding-bottom" ng-href="{{ideasView.parseTeamEmail()}}">
```
YashdalfTheGray: Did you get a chance to see the PR that I pulled in material-icons with? Either way, this is the code that you're looking for: 
```html
<md-button class="md-icon-button" ng-click="parseTeamEmail()" aria-label="Email Group">
    <i class="material-icons icon-button-adjust">&#xE0BE;</i>
    <md-tooltip md-direction="bottom">Email Team</md-tooltip>
</md-button>
```

Couple of things with the code snippet
* When you want a button, use `<md-button>`, that's what it's there for.
* There is a directive called `ng-click`, it's meant to handle clicks specifically. 
* We don't use the `<md-icon>` directive, it's only purpose is to take an svg icon and put it on the page but it has been replaced by `<i class="material-icons"></i>` since we started using the material icons font.
* You're going to want to read the Angular and the ngMaterial docs again.

Mctalian:
Please add an "Email Team" button to the team tab like we discussed at Peer Programming. Some people might expect an email button on the idea, but others might go to the team tab to contact the team.

src/ideas/ideasView/ideasView.js
```diff
@@ -398,6 +398,15 @@ angular.module('flintAndSteel')
                 }
             };
 
+            ctrl.parseTeamEmail = function parseTeamEmail() {
+                var emailString = "mailto:";
+                $scope.idea.team.forEach(function(teamElement) {
```
Mctalian: Since the idea needs to be retrieved from the server, there is a small window of time where `$scope.idea.team` is `undefined` and it causes console errors to appear (in the browser's JavaScript Console, not in the cmd prompt running the server). I think we need to add a check to keep these errors from being thrown:
```
if (angular.isDefined($scope.idea.team)) {
...
}
```

LongArmMcGee:
@YashdalfTheGray, I first started with code very similar to what you post here (minus the ng-click). But using md-button makes it look like shit. The button is way out in right field. Alan helped me get it to look better, but we must have forgot to test the click behavior : /

That's why I switched to md-icon. It looked a lot better b/c it was at the same high and stuck closer to the Idea Title text.

alanlgirard:
Apparently wasn't thinking straight that night!  @LongArmMcGee a simple fix to use the md-button like it was originally would be to put a layout tag of row in the h1 tag like:

            <h1 layout="row">{{idea.title}}
                <md-button class="md-icon-button" ng-href="{{ideasView.parseTeamEmail()}}" aria-label="Email Group">
                    <i class="material-icons icon-button-adjust">&#xE0BE;</i>
                    <md-tooltip md-direction="bottom">Email Team</md-tooltip>
                </md-button>
            </h1>

@YashdalfTheGray a ng-click will not work with how the parseTeamEmail() function is currently written so keeping it as a ng-href should be acceptable for the email team icon.

LongArmMcGee:
I can look at rewriting parseTeamEmail, but that's a different discussion.
I'll try the layout tag tonight.

Thanks Alan!

Mctalian:
Unless we can programmatically trigger "href-like" behavior from an
ng-click, I don't think you can write `parseTeamEmail` any differently.
`mailto:` only works in hrefs, so there'd have to be some angular service
that we can leverage in the controller to make this work. I don't know if
that exists, but it very well might. Or perhaps I'm misinterpreting the
problem altogether.

LongArmMcGee:
We have the same understanding of the problem. That was exactly why I used
the ng-href. I'm ignorant in the web dev world so I wasn't sure if there
was any other (better?) way to get the same behavior without using ng-href.

LongArmMcGee:
image
This results from adding the isDefined as below..
'<if (angular.isDefined($scope.idea.team)) {
                    var emailString = "mailto:";
                    $scope.idea.team.forEach(function(teamElement) {
                    // emailString + memberId.getProperty('email') + ';';
                    emailString += teamElement.member.mail + ';';
                    }
});>'

Is there a way to make it not call parseTeamEmail until it's clicked? I think that'd be a good solution if it's possible.

Mctalian:
Your screen shot, are there any browser console errors on that? I'd like to
take a look at something. Are you returning a blank string if team is
undefined or are you not returning anything? If it's the latter it's
probably choking on an undefined value in the href.

Mctalian:
Yeah if your code above is what you actually typed, then you're going to be returning `emailString` outside the `if (angular.isDefined($scope.idea.team)` block so `emailString` in that case in undefined. An undefined ng-href will very likely break things as you saw first hand.

```
...
var emailString = "";
if (angular.isDefined($scope.idea.team)) {
    emailString = "mailto:";
    $scope.idea.team.forEach(function(teamElement) {
        emailString += teamElement.member.mail + ';';
    }
}
return emailString;
...
```

Mctalian:
My suggestion above works when tested locally.

Also the email icon in the team tab, can it go above the team like the icons appear in the Backers tab? I like how that looks and I'd like to get a consistent look and feel on that front.

Mctalian:
This was approved by the team during Peer Programming.

PR #191, opened 2015-12-04
Add and edit search tags to an Idea

Daviddt17:
## What I Added:
I added the ability to add and edit search tags to an idea. Since chips in Angular Material are so broken, I opted to add tags using the custom chips I believe @alanlgirard added. There is a tag input in which every time the use presses enter, the tag will be added to the tag list which appears below the input. I've added the constraints of a limit of 5 tags and no duplicates.

Once Angular Material chips become stable it would probably be a good idea to switch over.

As of typing this I realized I didn't check for an empty tag, so I'll add that after I post the PR.

## What I changed:
I had to change the "editIdea" function as it did not take tags as a parameter. I also had to add tags to the projection of "getHeaders" so that the tags can be searchable. Hopefully neither of these changes will conflict with the RESTful changes @YashdalfTheGray is making.

## Some Screenshots:
addtags

ideatagsview

@YashdalfTheGray @bcbrennecke @Mctalian @alanlgirard @monotkate @LongArmMcGee @davethenipper

### DIFF
--- a/src/ideas/addIdeaView/addIdeaView.js
+++ b/src/ideas/addIdeaView/addIdeaView.js
@@ -24,7 +24,7 @@ angular.module('flintAndSteel')
             };
 
             $scope.addTag = function addTag(tag) {
-                if ($scope.idea.tags.length !== 5 && !$scope.doesTagExist(tag)) {
+                if ($scope.idea.tags.length !== 5 && !$scope.doesTagExist(tag) && tag !== '') {
                     $scope.idea.tags.push(tag);
                 }
             };
--- a/src/ideas/addIdeaView/addIdeaView.js
+++ b/src/ideas/addIdeaView/addIdeaView.js
@@ -13,11 +13,38 @@ angular.module('flintAndSteel')
             }
 
             $scope.idea = {};
+            $scope.idea.tags = [];
+            $scope.tagInput = "";
+
+            $scope.doesTagExist = function doesTagExist(tag) {
+                if ($scope.idea.tags.indexOf(tag) === -1) {
+                    return false;
+                }
+                return true;
+            };
+
+            $scope.addTag = function addTag(tag) {
+                if ($scope.idea.tags.length !== 5 && !$scope.doesTagExist(tag)) {
+                    $scope.idea.tags.push(tag);
+                }
+            };
+
+            $scope.tagKeyEvent = function tagKeyEvent(keyEvent) {
+                // Enter
+                if (keyEvent.keyCode === 13) {
+                    $scope.addTag($scope.tagInput);
+                    $scope.tagInput = "";
+                }
+            };
+
+            $scope.removeTag = function removeTag(tag) {
+                var index = $scope.idea.tags.indexOf(tag);
+                $scope.idea.tags.splice(index, 1);
+            };
 
             $scope.addNewIdea = function addNewIdea(ideaToAdd) {
                 ideaToAdd.authorId = loginSvc.getProperty('_id');
                 ideaToAdd.eventId = "";
-                ideaToAdd.tags = [];
                 ideaToAdd.rolesreq = [];
                 ideaSvc.postIdea($scope.idea, function postIdeaSuccess(data) {
                     if (angular.isDefined(data.status) && data.status === 'Created') {
--- a/src/ideas/addIdeaView/addIdeaView.js
+++ b/src/ideas/addIdeaView/addIdeaView.js
@@ -27,7 +27,7 @@ angular.module('flintAndSteel')
             $scope.addTag = function addTag(tag) {
                 var reNonAlpha = /[.,-\/#!$%\^&\*;:{}=\-_`~()<>\'\"@\[\]\|\\\?]/g;
                 tag = tag.replace(reNonAlpha, " ");
-                tag = _.camelCase(tag);
+                tag = _.capitalize(_.camelCase(tag));
                 if ($scope.idea.tags.length !== 5 && !$scope.doesTagExist(tag) && tag !== '') {                    
                     $scope.idea.tags.push(tag);
                 }
--- a/src/ideas/addIdeaView/addIdeaView.js
+++ b/src/ideas/addIdeaView/addIdeaView.js
@@ -1,4 +1,5 @@
 /* global angular */
+/* global _ */
 
 angular.module('flintAndSteel')
 .controller('AddIdeaViewCtrl',
@@ -24,7 +25,10 @@ angular.module('flintAndSteel')
             };
 
             $scope.addTag = function addTag(tag) {
-                if ($scope.idea.tags.length !== 5 && !$scope.doesTagExist(tag) && tag !== '') {
+                var reNonAlpha = /[.,-\/#!$%\^&\*;:{}=\-_`~()<>\'\"@\[\]\|\\\?]/g;
+                tag = tag.replace(reNonAlpha, " ");
+                tag = _.camelCase(tag);
+                if ($scope.idea.tags.length !== 5 && !$scope.doesTagExist(tag) && tag !== '') {                    
                     $scope.idea.tags.push(tag);
                 }
             };
--- a/src/ideas/addIdeaView/addIdeaView.tpl.html
+++ b/src/ideas/addIdeaView/addIdeaView.tpl.html
@@ -12,7 +12,10 @@
             </md-input-container>
             <md-input-container class="idea-tags-input-container">
                 <label>Tags</label>
-                <input ng-model="tagInput" ng-keypress="tagKeyEvent($event)" type="text"></input>
+                <input name="tags" ng-model="tagInput" ng-keypress="tagKeyEvent($event)" maxLength="100" type="text"></input>
+                <div ng-messages="tags.$error">
+                    <div ng-message="" class="tag-message">Press Enter to Add Tag - Limit 5</div>
+                </div>
             </md-input-container>
             <div layout="row">
                 <span class="custom-chip-wrapper" ng-repeat="tag in idea.tags">
--- a/src/ideas/addIdeaView/addIdeaView.tpl.html
+++ b/src/ideas/addIdeaView/addIdeaView.tpl.html
@@ -12,7 +12,7 @@
             </md-input-container>
             <md-input-container class="idea-tags-input-container">
                 <label>Tags</label>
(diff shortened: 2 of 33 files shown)


Daviddt17:
Added blank tag check and tests.

YashdalfTheGray:
Yeah, they do. I'm at least touching (if not changing) 100% of the server code and the services layer code on the client side. 

It's all good though. Do you care if we wait to merge this till my delivery? I don't want to force you to handle the merge conflicts though, so if you don't want to, you can merge this earlier.

Mctalian:
I haven't looked at it yet, but since his is smaller, is only touching a
bit of the server code,  and is nearly ready now, I'm kind of leaning
towards merging this before the server backend changes because those could
take a while and I'd hate to hold up progress if they take longer than
expected.

YashdalfTheGray:
That's cool with me!

Daviddt17:
I'm fine with whatever. I don't mind having to deal with merge conflicts if I have to, but if we think moving forward with this PR first is the better choice then just let me know and I'll go ahead and merge :)

Mctalian:
It may not be obvious for users to hit "enter" to submit their tags and it could result in tags not being submitted. This also occurs for a new idea:

tags

Also, allowing spaces and commas should probably not be allowed. Can you do a `.split(',')` or something like that to allow the tags to be submitted in a comma-seperated list? Replace spaces with dashes, trim off trailing spaces, etc. Up to you how you want to handle this. That may be the first thing that people try to do, I'm not sure. In other sites with keyword fields, I usually assume comma or semi-colon delimited lists. It's too bad md-chips doesn't work correctly. That'd probably fix this. We might also want to provide some help text like "Limit to 5" or at least "Press 'Enter' to add tags" if we can't find a good way to work this. Maybe pop up a warning if users try to submit or edit an idea with text in the tags field? Perhaps we automagically assume they want those tags? There's a lot you can do here.

image

LongArmMcGee:
what do you use to make those gifs booby

monotkate:
I feel like people could figure out that enter submits after a try or two.  You can edit if they're issues, so while it may take a second, I think it would be fine.  Limiting the number of tags and nixing commas (or punctuation in general really) I agree with

Mctalian:
It's called ScreenToGif it's pretty neat.

YashdalfTheGray:
I agree that people will figure it out after a second or two. Keep it.

Yash Kulshrestha
yash.kulshrestha@gmail.com
(614)563-1163

Daviddt17:
Sorry for not getting back to this until today. 

So what should be done when a user enters any special character other than a space. I was thinking show a warning message. The angular material website uses ng-messages for this, which I'm not sure if it built in to material or not. Do I need to include it? Should I?

YashdalfTheGray:
`ng-messages` is built into angular and ngMaterial has support of them too. You should be good to go on using them!

Mctalian:
You might be able to "split" based on a regex for punctuation, or ignore it like we do in searches. Maybe split on ", " (comma followed by a space) and each of those elements would be considered a new element in the tag array. If that makes sense.

Warnings for non-comma punctuations (maybe also non-semicolon if we want to allow those) are probably a good thing. Maybe allow commas and semicolons and do the split like I mentioned in the paragraph above.

Mctalian:
Personally I don't like the "they'll figure it out in a second or two" approach. Because it won't be a second, or two, it'll be an attempt or two (at the very least, possibly worse. Some people just give up after things don't work). Can we use the submit button to act like what the "Enter" button does?

If not, I guess just leave it. Just seems like we're being lazy from a UX perspective here to just leave the user to figure things out for themselves.

monotkate:
See, the reason I lean towards "they can figure it out" is because I feel like I've had to do that on a lot of sites, particularly when learning to tag things.  I could also see the argument that if we're trying to work with an older generation they may not be as apt to pick it up, but those are likely not the people who are going to be using the product.  We could have some sort of official discussion about this though.

YashdalfTheGray:
That's exactly what I've been trying to say all this time. I don't think that we should design for the older generation, we should design for the people who are actually going to use this product. Granted there can always be improvements on the UX side but the group that we're targeting is already familiar with backing a project, tagging things on the internet, common iconography, etc. That's our target group. That's what the design should be based on.

Mctalian:
I still disagree, but we'll have a discussion at another time. I think my
other suggestions on how to handle punctuation and spaces could still work
despite this other discussion.

Daviddt17:
Right now this is what I see as expected behavior:

1. Cannot enter blank tag
2. Cannot enter tag with special characters (excluding space, comma, and semi-colon)
3. Remove leading and trailing spaces
4. Replace inner space or repeating spaces with single dash
5. Allow user to enter multiple tags at once using ', ' or ';' as delimiters

monotkate:
@Mctalian what if we add a little info hover/click button that gives some instructions for tagging.  It shouldn't interrupt the design or functionality being simplistic, but for those that can't get it you can get some more info about making it work.

Daviddt17:
If we accept number 5, what should we do for repeated commas or semi-colons? Treat them as a single comma or semi-colon? Don't allow the user to enter?

YashdalfTheGray:
[Use it well](https://lodash.com/docs#camelCase) and remember, with great power comes great responsibility! :smile:

Daviddt17:
I think I made the mistake of pushing my changes before rebasing. That's why there are duplicate commits... I think... maybe?... I suck at git.

Anyway, I made a few simple changes. All special characters are ignored and the tag is changed to camelCase. I also added the message under the input field like so:

tagstip

I'm not sure if I went about the message the right way. Any feedback?

Daviddt17:
I should be a little more specific. Special characters are replaced with spaces and then camelCase is applied.

Daviddt17:
This is ready for review. @YashdalfTheGray @Mctalian @LongArmMcGee @monotkate @alanlgirard @davethenipper @bcbrennecke

Mctalian:
I like it. Approved.

Funny story: please do not make any changes based on it!

I happened to find myself on one of those "23 of the Funniest Things on Tumblr that are GUARANTEED to make you laugh EVERYTIME!!!!!111one!!" posts on Facebook. And while they certainly did NOT make me laugh, not even once, I did see something interesting that DID make me laugh and cry simultaneously:

image

Yes... they allow spaces in their tags. My whole world and existence is a lie.

Daviddt17:
"#4 will change your life"

One note about not having spaces. Since we don't remove spaces in our normalized search fields, you have to search for tags without spaces in the search. For example,  if I want to search the tag "thisIsATag" I can't find it by typing "this is a tag". It might be nice to remove spaces everywhere to make searching easier.

Daviddt17:
Haha I didn't mean to tag issue number 4

Mctalian:
But #4 DID change our lives!

I actually did have that thought when I saw the spaces in the Tumblr tags. Maybe instead we can modify searching to split the query into an array of strings separated by spaces and then the search would return ideas that matched all of these. It would also help with people typing words slightly out of order with what is in the abstract or idea title. Removing spaces would require people to type phrases perfectly in order (the way it is now, so it's probably an ok change for now). Not sure if you want to include this in this PR or make a small one after. Depends on what the card covers, I guess. Your choice. Or if anyone else has a comment on this.

Daviddt17:
Alright I added the space to the search regex. I also added a ng-disabled to the tag inputs when 5 tags is reached.

src/ideas/addIdeaView/addIdeaView.js
```diff
@@ -13,11 +14,41 @@ angular.module('flintAndSteel')
...
+                if ($scope.idea.tags.indexOf(tag) === -1) {
+                    return false;
+                }
+                return true;
+            };
+
+            $scope.addTag = function addTag(tag) {
+                var reNonAlpha = /[.,-\/#!$%\^&\*;:{}=\-_`~()<>\'\"@\[\]\|\\\?]/g;
+                tag = tag.replace(reNonAlpha, " ");
+                tag = _.camelCase(tag);
```
YashdalfTheGray: If you throw a `_.capitalize()` in front of this, it'll make it so that the tags come out looking like "#ThisIsATag". So `_.capitalize(_.camelCase(tag));`, I don't know if I was clear on where capitalize should go.

src/ideas/addIdeaView/addIdeaView.tpl.html
```diff
@@ -1,21 +1,34 @@
...
+            </md-input-container>
+            <md-input-container flex>
+                <label>Description</label>
+                <textarea ng-model="idea.description" columns="1" md-maxlength="1000" required></textarea>
+            </md-input-container>
+            <md-input-container class="idea-tags-input-container">
+                <label>Tags</label>
+                <input name="tags" ng-model="tagInput" ng-keypress="tagKeyEvent($event)" maxLength="100" type="text" ng-disabled="idea.tags.length === 5"></input>
+                <div ng-messages="tags.$error">
+                    <div ng-message="" class="tag-message">Press Enter to Add Tag - Limit 5</div>
```
YashdalfTheGray: Yeah, I think you did `ng-messages` right!

YashdalfTheGray:
Yeah, I don't know how git's going to manage these duplicate commits but whatever, they all seem to have different hashes. Are you using github for desktop? Use the command line instead, I don't trust that app. 

So spaces in our tags is a thing or can you find a tag by using the tag text with spaces? Because spaces in a tag is an absolute NO NO and I won't be pressured by tublr or anyone else who uses that shitty way of hashtagging. 

Unless you already have the functionality to get multiple tags out of a comma separated list, just check in the enter functionality right now, if it becomes really really necessary, we can change that later.

Daviddt17:
Nope I use command line. I think it was because I pushed my changes, synced with master, then pushed again. I should have have just committed my changes, synced with master, then pushed. Not really sure though.

All spaces are removed from the tag. I did add a space to the search regex which will remove all spaces from the search fields. This lets us search "This is a tag" and will return ideas with the tag "ThisIsATag".

I'll add the capitalize now.

Mctalian:
Approved!

Mctalian:
Can we get another approval? @YashdalfTheGray @alanlgirard @davethenipper @bcbrennecke @monotkate @LongArmMcGee Bueller? 
image

alanlgirard:
Approved.  I don't think the message will show either but I can address that in my PR #197 easily

Daviddt17:
Sounds good. Merged!

PR #192, opened 2015-12-09
Homepage Redesign

bcbrennecke:
Alright I wanted to get a pull request in so that everyone can check out the layout and get some suggestions and comments in before pair coding this week.  There is definitely a lot of work left to do.

I need thoughts about the concept and additions to be made.  The code is still a mess and needs refactoring, so don't waste time parsing through that.  One comment about tripling the css file size and i'll go nuts.  I'll refactor before an actual merge.

Things I know are wrong and I need help on:
-Log In button is not linked to anything
-Inno/Support/Comment/Team cards should not push other content down when they fold out.
-We need some of our own pictures for the background
-Text color for 'Search Ideas' needs changed

Let's hear some feedback! Go!

@YashdalfTheGray @alanlgirard @Mctalian @LongArmMcGee @monotkate


monotkate:
- text for innovation challenge ends up being black on dark.  Not readable
- search ideas has a similar problem.  I like the concept, not sure on the coloring
- I really like the tips for various things, but I would recommend having Log In consistently as the first option.  It really should be, particularly when you're giving directions
- This is nitpicky, but I don't like how the + idea button formats the text.  Can you shrink it for display purposes there?
- LOVE the timeline.  Maybe add years?  Seems assumed, but when you get to the end of the year it's a little hard to tell what's what
- Probably a down the road feature, but potentially make the advice boxes on what you can do actually link to actions.  Like, "idea" links to the "all ideas" page.
- Minor, I would use the chat icon for comments instead of people, since that's used for team
- I'm assuming the timeline will eventually link to events once we add events?  Because that would be awesome.
- I can help with the login button since I set that all up in the toolbar.  Let me look at it this morning when the kids are in lab
- Overall I really like the redesign, I think it does a lot for the page.

monotkate:
image
Actually, here's the code for the login button.  I added the person icon to try to imitate toolbar login and the tooltip.  Which is kind of the same as the button, but links it to all the other buttons.  

One thought I had was actually removing the login step if the user is already logged in.  Otherwise it seems like you're telling them to do something they've already done.  Just slap an ng-show on it

monotkate:
Whoops, sorry, saw you mentioned the search ideas color already.  
Recommendation: Make it grey or white, not red.  Just bump the opacity up to like .8 or .9, and then add a break underneath so that it bumps it up from the bottom.  It will look a lot more fluid with the rest of the interface.  It looks really really good when you do that imo.  Awesome job

bcbrennecke:
This is perfect.  Yeah the idea is that events will all link to an event page which will display the problems posted for that event.  I am also going to add a bar right before the timeline called 'Active Events' So they're right there and easy to see and click on for what's current.
Definitely colors is a problem, I wasn't sure what kinds of pictures we would do in the background but yeah i'll change those to lighter colors.  
Good idea with the login step, that should be easy.
Knew i'd have to change the people pic but wasn't sure what too, good call with chat icon.
Any other content you can think of that would be useful right on the homepage?

Mctalian:
I like what I see. Here are a few additional (hopefully not contradictory) suggestions from above:
* Perhaps show the ideas when the cursor is active in the search bar or if there is text in the search bar
    * For instance:
        1. User Clicks on Search Bar, ideas show up (`ng-focus`)
        2. User types search text, ideas are filtered
        3. User clicks out of the search bar, ideas remain
        4. User clicks on search bar again and removes all text, ideas remain
        5. User clicks off the search bar, ideas disappear and FAQ tiles reappear (`ng-blur`)
* Maybe hide the timeline when the ideas are showing as well?
* Perhaps add animation to the FAQ tiles on hover (this might be tough or impossible depending on how it's implemented. Animation on Angular confuses me so if we can't do this it's fine). For whatever reason it throws me off when I hover over and everything moves. Could be the concussion, though.
* Alternatively, can we make it so the 4 tiles fit across the screen on big screens and adjust responsively as the screen shrinks? 2 x 2 as we have it now on Medium screens? All in a column on small screens and mobile? It actually works fairly well on mobile already, but having the tiles fit the width of the screen when it's that small would be best (they all are different sizes and stick to the left side of the screen, but expand very nicely when tapped). It would also help people realize the Events are down there. It's hovering right around "the fold" so I honestly didn't even see them until I typed in the search box.
* Maybe under "Support" you can offer "clicking the heart" as an alternative method of support. Something like "click the heart to simply like the idea, or head to the backers tab if you have more to offer!" I can do without this one, just noticed we covered almost every action a user can perform on an idea, except likes.

Mctalian:
Oh and I can't believe you increased the size of the CSS file that much! It's somewhere between doubled and quadrupled! Crazy!!!

In other news, we might need to look into a different version of the timeline stuff. I suspect your TravisCI build failure is because the timeline is requiring some version of angular than the one we're specifying and the one we're specifying is not compatible with timeline so bower flips out, throws a tantrum and cries. I can help you with this whenever you're up for it.

monotkate:
I'm not sure I like the idea of ideas superseding what's on the screen.  If that's the case we can guide them to the all ideas page in order to search.  

Agree with responsive tiles though.  We had a chat about making them smaller as well though.  Could potentially drop down some sort of tip or slide down some sort of tips list instead of messing with the layout?

alanlgirard:
Overall like the new direction with the homepage.  Have a couple of suggestions

* Idea icon on Innovate card should probably have the mouse-over pointer removed unless we make this actually link to create a new idea.
* Log in should probably be the first step of all of the cards.
* To prevent the resizing/layout of the cards you could use a fixed container and have the icons shift to the left revealing the text on the right on mouseover inside of the container
* To not have to worry about the Innovation Challenge text color conflicting with the picture, could possibly move it above the picture.  This would allow the picture to be more prominent especially if we end up using pictures of innovation participants etc
* With the more vertical layout of the homepage, we might want to utilize anchors with some page navigation so people can quickly jump to the event schedule etc

Note to self...stop using etc, etc!

src/assets/app.css
```diff
@@ -203,6 +203,328 @@ body {
...
+                    }
+
+                    .timeline-centered .timeline-entry .timeline-entry-inner .timeline-label h2 span {
+                        -webkit-opacity: .6;
+                        -moz-opacity: .6;
+                        opacity: .6;
+                        -ms-filter: alpha(opacity=60);
+                        filter: alpha(opacity=60);
+                    }
+<<<<<<< HEAD
```
YashdalfTheGray: I don't know if this is valid CSS :smile: 

Make sure you don't check in merge markers.

src/homeView/homeView.js
```diff
@@ -3,13 +3,56 @@
...
+        '$scope', '$state','$mdSidenav', 'ideaSvc',
+        function($scope, $state, $mdSidenav, ideaSvc) {
             "use strict";
 
             $scope.navToBrowse = function navToBrowse() {
                 $state.go('ideabrowse');
             };
+
+            $scope.hoverInLeft = function(){
+                this.hoverLeft = true;
```
Mctalian: I would recommend something like
``` JavaScript
var ctrl = this;
```
At the top of the file because `this` gets a little tricky inside function scope especially as you start nesting. This way you always know which `this` you're referring to.

src/homeView/homeView.tpl.html
```diff
@@ -1,24 +1,233 @@
...
+        <div>
+            <h1>Innovation Challenge</h1>
+        </div>
+    </div>
+    <div flex  layout-align="center center" >
+        <div layout="column" style="width:100%" layout-align="center center">
+            <div class="homeview-login-button">
+                <md-button class="md-raised" >Log In</md-button>
+            </div>
+            <md-input-container class="homeview-search md-accent md-hue-1" flex>
```
Mctalian: If you remove the `flex` on this line it should make it compatible with IE. At least this particular search box. Unsure about the rest of the page.

PR #195, opened 2015-12-10
Fixes botched merge from a LONG time ago

### LINKED ISSUE #194: Bad merge, update deleting
335f0872a7ec25a4e2626e9731777d74de6b6e02

There may be more in this merge that I missed. We are using $index based deleting now instead of just passing in the object that we want to delete. It gets confusing if we sort the info in a different way, so it's better to just pass in what we want to delete, not its potentially changing position.

Mctalian:
Closes #194, see commit comments or issue discussion for a bit more info.


Mctalian:
@YashdalfTheGray , @monotkate , @alanlgirard 

I guess since this fixes functionality I could just merge it, but considering I'm fixing my own mistake here, I'd like to get an eye or two on this.

YashdalfTheGray:
Yeah, this looks right to me. Approved!

monotkate:
Approved!

Issue #153, opened 2015-11-19
Length of idea title


bcbrennecke:
I can create an idea with no title.  Then when you delete it it is deleting "null".

Mctalian:
Look into:
[Angular Input Parameters like ng-minlength](https://docs.angularjs.org/api/ng/input/input%5Btext%5D)
[Angular Material Input Parameters like md-maxlength](https://material.angularjs.org/latest/api/directive/mdInput)

Ultimately we need to start adding "required" fields to our forms. It will knock out a lot of issues like this.

alanlgirard:
Closed with PR #197

Issue #179, opened 2015-12-02
Character Limit On Forms


davethenipper:
Every single form on the website should have a minimum (four) and maximum character limit.  This should address a few other bugs.

alanlgirard:
Closed with PR #197

Issue #177, opened 2015-12-02
Unknown Comment Character Limit


davethenipper:
The user does not know what the comment character limit is, or when it has been reached.

alanlgirard:
Closed with PR #197

Issue #176, opened 2015-12-02
You Can Attempt a Login with Blank Username/Password


davethenipper:
Users are able to attempt to login with a blank username/password.

Discussion has been: Don't allow them and use an X?

YashdalfTheGray:
Attempt but it's not successful, yes? When I worked on login, you could try it with blanks and it would just say that those fields are required. 

If this is still true, then it is the intended behavior, no need to worry about it.

Mctalian:
We have a lot of issues with form validation (hence the description issue).
I am of the camp that if something is required, don't even let them submit
while it's blank. Disable the button. I think Login should work this way as
well.

YashdalfTheGray:
We can disable the login button, sure or we can use the `ng-required` and the `<ng-messages>` directive to make sure that we give user feedback on crappy input and the disabling button logic comes built in. They could sit there and click the button all day and nothing would happen provided that all of this sits inside a `<form></form` element.

Mctalian:
I don't think disabling button logic comes for free. I've always had to utilize `ng-disable` and watch for the form to be `$valid` or whatever. I could have been doing it wrong, but I don't want us to make an assumption that something might come for free, if it might not. To be confirmed.

YashdalfTheGray:
That's a good way to do it. If you use `ng-required` and watch for the `$changed` event and see if it's valid. Form validation comes for free if you use `ng-required` and the `<ng-messages>` directives. It will let you click the button but won't run anything because of the `ng-required` checks not passing.

alanlgirard:
Closed with PR #197

PR #197, opened 2015-12-11
ngMessages/Form Validation

alanlgirard:
Added ngMessages dependency so that our ng-messages show properly.  Now if you want multiple messages they will populate properly based on the criteria such as minlength or required, etc.

Added messages/form validation to both the login and the addIdea.  I limited the addIdea title to 4-30 characters and the description to 15-1000 characters.  I am open to suggestions if we want to alter some of these or get rid of them all together?

~~Re-factored the account page to display pertinent info.  Only implemented a "My Ideas: tab currently but left in place holders for "My Backs" and "My Teams" once the back-end is done being edited.  Right now there isn't an easy way to grab a user's backs/teams without looping through all of the ideas and checking.~~

@monotkate @Mctalian @YashdalfTheGray


src/ideas/addIdeaView/addIdeaView.tpl.html
```diff
@@ -1,19 +1,40 @@
...
+                              md-maxlength="1000"
+                              ng-required></textarea>
+                    <div ng-messages="newIdeaForm.desc.$error" role="alert">
+                        <div ng-message="required">Required</div>
+                        <div ng-message="minlength">Your description is too short. Must be at least 15 characters</div>
+                        <div ng-message="maxlength">Your description is too long</div>
+                    </div>
                 </md-input-container>
-                <p></p>
+                </p></p>
```
Mctalian: Why 2 closing tags? (To be honest I never liked this random `<p></p>` to begin with, but why is it now `</p></p>`?

Can we use padding instead?

Mctalian:
If you edit the idea and exceed the 1000 character limit, you can still submit the edits and the description disappears.

src/users/accountView/accountView.tpl.html
```diff
@@ -1,12 +1,47 @@
...
-        <h2><center>{{user.name}} (<i>{{user.username}}</i>)</center></h2>
-        <h4><center>{{user.email}}</center><h4>
-    </div>
-    <md-button class="md-raised md-warn" ng-click="logout()" flex>Logout</md-button>
-</md-content>
+    </md-content>
+</md-card>
+<md-card>
+    <md-content>
+        <md-tabs md-dynamic-height md-border-bottom="">
```
Mctalian: I'm not really feeling the use of tabs here. I think we should try to follow more of a GitHub pattern and have different partitions on the screen where it shows "top 5" ideas, and maybe 5 most recent backs in another partition, and a "liked ideas" partition that will show the 5 "most popular" ideas that you've liked.

Some of this is more "future", and maybe we don't care for this commit. Or maybe people disagree and like the tabs, if that's the case, I'd like the ideas to not be listed straight down and look more like the All Ideas page instead.

davethenipper: That's a neat idea.  Some of the tabs are starting to get a little
overwhelming.

src/ideas/addIdeaView/addIdeaView.tpl.html
alanlgirard: Left-over from messing around.  Removed and added padding in instead.

src/users/accountView/accountView.tpl.html
alanlgirard: Yeah I didn't really like the tabs either but couldn't think of a good way to lay it out.  I do like the github profile page and will work on re-factoring the account page this weekend to have more of a partition look

alanlgirard:
Reverted the account page for the time being while I work on it.  Want to get the ngMessages and form validation through first to not hold up anyone at all.  This PR should be good to be approved for the use of ngMessages on forms.

@Mctalian @monotkate @davethenipper @YashdalfTheGray

monotkate:
I love the ellipses on a long title for the sidebar nav.  :100: 

On login, I kind of liked the required notifications for the color, but I suppose there's also assumptions on login.

For add/edit idea I'm a fan of adding the character limit either when you're clicking on the input itself, or always.  I don't like that it doesn't show things until you click on and then off.  It means you basically have no idea what you can input up to until you screw it up :(

alanlgirard:
There is an issue with md-maxlength and ng-messages where the character count gets put in the ng-message (https://github.com/angular/material/issues/5853). I was hiding the message unless there was an error.  To fix this for the time being I changed the messages to show all of the time for the addIdea and editIdea pages.

Mctalian:
Will this close any of the existing issues? Or does it only partially address one/some?

Can you tag appropriate issues?

alanlgirard:
Closes issues #153, #175, #176, #177, and #179

Mctalian:
Thank you for doing that.

I don't see any changes that would clear up #177, what is the character limit for comments?
And as far as #179, what about comments, backs, and updates?

YashdalfTheGray:
I like this! Approved!

alanlgirard:
Added validation for Updates, Comments, and Back description.  All are limited to 4-500 characters.  The back description is still not required but if they enter any text then it has to be within 4-500 for them to submit.

Should now fully close out #177 and #179 unless there are any other forms that I have missed.

monotkate:
Approved!

Mctalian:
:+1: Nice work!

Mctalian:
a.k.a Approved

PR #200, opened 2015-12-11
Indicates which Login Credentials

### LINKED ISSUE #178: No Login Credential Indication
The user does not know what login credentials to use.  We should inform them it supports their single sign on.

Side note: Saying that we store their credentials is wrong.

Mctalian:
Closes #178

There is now a bit of help text that lets users know which credentials to use when logging in.

image

### DIFF
--- a/src/assets/app.css
+++ b/src/assets/app.css
@@ -26,6 +26,13 @@ body {
     margin-right: 10px;
 }
 
+.sign-in-tip {
+    font-size: 12px;
+    padding-top: 10px;
+    margin-left: auto;
+    margin-right: auto;
+}
+
 .signup-button {
     margin-left: 10px;
 }
@@ -122,6 +129,14 @@ body {
 
 .extra-padding-left-md {
     padding-left: 10px;
+}
+
+.extra-padding-top-md {
+    padding-top: 10px;
+}
+
+.extra-padding-15 {
+    padding: 15px;
 }
 
 .extra-padding-lg {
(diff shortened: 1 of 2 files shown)


Mctalian:
@bcbrennecke approves, he walked by moments after I put up this PR

monotkate:
personally I would adjust to say something about Rockwell credentials or something, but otherwise windows would work

LongArmMcGee:
Looks reasonable

alanlgirard:
I don't know if I would word it as Windows Credentials.  I know there are some macs in the company(MFT) so it is probably better to use Rockwell or RA-INT in the description.

Mctalian:
Rockwell uses at least 2 logins. So even that is misleading unfortunately.

monotkate:
So basically there's no good way to clarify? lol.  What are we drawing from for login then?  Which Rockwell login?

YashdalfTheGray:
How about "Use your Rockwell SSO"? I know for a fact that most people know what that is because a recent training mentioned something like that.

LongArmMcGee:
Never heard of SSO. I would assume it means single sign on? Wouldn't
'RA-Int Domain' be the most descriptive?

Mctalian:
RA-INT is probably the best we're going to get because single sign on to me
could also be the SharePoint login (ADFS).

LongArmMcGee:
Settled.

YashdalfTheGray:
Approved!

Mctalian:
Verbal approvals from @alanlgirard and @bcbrennecke. Plus @YashdalfTheGray's approval above. I'm merging.

PR #201, opened 2015-12-11
Icon Change and Idea in Browse Update

### LINKED ISSUE #186: Time Created vs. Time Modified
Time created and time modified need to be consistent throughout the platform.

monotkate:
Changed the icon for backing.  Opinions welcome.  Mostly because I wanted to use the backing icon for team.  Added the team count in an idea page.

Also changed text on the ideas in browse from text to icons with tooltips and shrunk author a bit.  I think it simplifies and looks a lot better.  I plan on attaching them to the "hasBacked/Liked/isinTeam" functions later, but they're not part of ideaSvc and I have to figure out how to swing that first.  

Uncommented the delete button but I realized there are some discrepancies between the data in the front and and what the backend is supposed to have.  It doesn't do anything right now and I can comment it back up.  I'm hoping to get a look at these discrepancies this weekend because @alanlgirard and I pinpointed a bunch of them.

Minor change overall.  Unrelated to the issues I've got in my name for now which are both affected by the discrepancies.  @YashdalfTheGray @Mctalian @alanlgirard 

EDIT: Fixed some inconsistencies --
- Backers weren't actually set up to use the time modified/time created.  It should be working correctly now from what I can tell.  
- Set most idea tooltips to scroll down instead of the team tab scrolling up and the backers tab scrolling down.  
- Updates can't be edited right now (I believe), but their time was also set up incorrectly using time instead of timeModified.  Should be fixed now.  Fixes #186 I believe

EDIT2 (Because no one's checking, so I just keep adding things):
- Changed out the idea icons interact.  Now depending on what you should be doing with backers/teams it will focus on that tab. Backer indicator will push you to backer tab. Team indicator will push you to team tab unless you are logged in and have not backed, at which point it will push you to the backers tab
- Deleted the dialog info for showing who has liked the idea and the associated test.  Can put back if necessary but right now it's not being used anywhere so I nixed the code.

### DIFF
--- a/src/ideas/ideasView/ideasView.js
+++ b/src/ideas/ideasView/ideasView.js
@@ -418,21 +418,6 @@ angular.module('flintAndSteel')
                 return false;
             };
 
-            $scope.showEditTeam = function(ev) {
-                var teamObj = '';
-
-                $mdDialog.show({
-                    controller: 'DialogBackCtrl',
-                    templateUrl: 'ideas/ideaTeam/ideaEditTeam.tpl.html',
-                    parent: angular.element(document.body),
-                    targetEvent: ev,
-                    clickOutsideToClose: true,
-                    locals: {
-                        backingObj: null
-                    }
-                })
-            };
-
             ///////////////////////
             // BACKING FUNCTIONS //
             ///////////////////////
--- a/src/ideas/ideasView/ideasView.js
+++ b/src/ideas/ideasView/ideasView.js
@@ -483,6 +483,7 @@ angular.module('flintAndSteel')
                             text: ctrl.editBackText,
                             authorId: back.authorId,
                             timeCreated: back.timeCreated,
+                            timeModified: back.timeModified,
                             types: $scope.selectedTypes
                         };
                     }
--- a/src/ideas/ideasView/ideasView.js
+++ b/src/ideas/ideasView/ideasView.js
@@ -212,7 +212,8 @@ angular.module('flintAndSteel')
                     obj = {
                         text: ctrl.newBack,
                         authorId: loginSvc.getProperty('_id'),
-                        time: now,
+                        timeCreated: now,
+                        timeModified: '',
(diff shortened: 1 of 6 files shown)


LongArmMcGee:
* I like the change to 'contributors' in the # of backers tool tip. 'Contribute' as a noun and verb are more descriptive than 'back'. I think the text throughout the site should reflect this change. IE: 'Backers' tab should now be 'Contributors' tab
* Since the # of likes icon and # of contributors icons are both actionable (toggle liking the idea, edit contribution) I think we should continue the trend and make the # of team members icon bring us to the team tab. or something...
* The tooltip for the like button is hanging a bit lower than the others
* I'm not sure if i like the new # of contributors icon, it looks like something that would be used to tag/label/or otherwise collate information. I'm not sure what else would work... maybe [these numbered icons?](https://www.google.com/design/icons/#ic_filter_1). [work icon?](https://www.google.com/design/icons/#ic_work)
* The smaller author name looks sooo much better. Something about those cards always bothered me.
* Icons on the idea browse cards are nice to keep the info presented consistently throughout the site

monotkate:
The goal for the backer icon was to find one that has a filled and empty icon like the other ones. We struggled for backs

Also, I believe there's a plan to rework the edit team thing to a dialog much like the backers. We can tie it into that pretty easily. Either me or @algirard will do the change.

src/ideas/ideasView/ideasView.js
```diff
@@ -506,8 +482,7 @@ angular.module('flintAndSteel')
                         newBack = {
                             text: ctrl.editBackText,
                             authorId: back.authorId,
-                            time: back.time,
-                            timeModified: now,
```
davethenipper: `back.timeModified`.  To preserve stuff in case it's backed twice quickly.

davethenipper:
Team Approved!

Mctalian:
Should close #186

PR #202, opened 2015-12-14
Fixes Edit Back, Enables Delete Back

### LINKED ISSUE #199: Someone broke hasUserBacked()
Not the biggest deal, but in the past hasUserBacked() returned `true` for the idea owner as well as the regular backers.  As of this morning (something was changed in the last night time frame because I rebased yesterday afternoon and it still worked), it no longer returns true for the idea owner, but it does for anyone else.  If this is intended, I can easily change the backing code to work with it, but I want to know if it's intended and why it changed.  @YashdalfTheGray @Mctalian was this either of you? @alanlgirard caught it.

Mctalian:
Closes #199

Also deleting backs now work @monotkate, you nearly had it. You had `back` as the interactionType not `backs`. I tried to think of how we can make that better but gave up for now. Sorry for the confusion!

There is a bit of quirky behavior with the backing dialog. When you click in between line items it will implicitly check the backing type, but not display the checkmark. It's a bit odd. I am going to look into this.

In the meantime, @YashdalfTheGray, how much do these changes mess up your backend stuff?

### DIFF
--- a/server/db.js
+++ b/server/db.js
@@ -140,6 +140,10 @@ module.exports = function(dbName, cb) {
                     console.error(chalk.bgRed(err));
                     cb(err);
                 }
+                else if (results.nMatched === 0) {
+                    console.error(chalk.yellow('No documents with id ' + id + ' were found.'));
+                    cb(null, results);
+                }
                 else {
                     // console.log(chalk.bgGreen('Document with id %s updated in the ' + collection + ' colletion.'), id);
                     cb(null, results);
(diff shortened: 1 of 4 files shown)


monotkate:
GODDAMMIT I WAS SO CLOSE!!!!

Approved.  Also, sulking.

monotkate:
Oh, I noticed that when I was messing with some other stuff.  I can look into it since I'm messing with some more dialog stuff anyway.  I'm also working on editing the owner back on a new idea.  It's acting weird with it.

Mctalian:
It's a simple fix, actually. Want me to just add it, it's a single line change?

I was looking at refactoring the dialog a bit. I think there's a simpler solution without nested for-loops, but it's one of those "it works just fine" things right now so I'm going to leave it.

monotkate:
Go for it

Mctalian:
Waiting on @YashdalfTheGray, I don't want to screw up your stuff.

If you approve, you can just merge it in.

YashdalfTheGray:
This is good, some of this functionality I had put into my code already but it's good that you got that and more! Approved, merging now.

Issue #183, opened 2015-12-02
Logging In Is Not Intuitive


davethenipper:
Some users may not know they need to log in to do things on the website.

Discussion: Grey out areas for those who have not logged in instead of hiding them.  Then make clicking these take them to the login page.  A good example is the create idea button.

monotkate:
I'm adding tooltips right now which should help with some. I think it also tells you to login to comment now. Unless I didn't end up adding it last time. Can we add something on the homepage to get users to login?

Daviddt17:
Yep, I second the login on the homepage. Even better, let's get rid of the default person icon when no one is logged in and replace it with a nice and shiny login button.

Daviddt17:
For instance, I think this:
![signin](https://cloud.githubusercontent.com/assets/5295379/11534727/2bd9e66a-98de-11e5-9802-e85dd86bf038.JPG)

Looks much better than this:
![personsignin](https://cloud.githubusercontent.com/assets/5295379/11534728/2e60db1e-98de-11e5-8d77-5e5f19765916.JPG)

It's way more obvious that it's a sing in button that way.

Mctalian:
I also think we can be a bit more "modern" with our homepage. I can provide
some examples later.

YashdalfTheGray:
@Daviddt17 that sign in button is against the material design guidelines. We can have text but not the red background. Also, Miranda is adding tooltips that should help. All apps should invite people to click around, they learn by doing, telling the user exactly what to do is bad app design. 

That being said, I agree that the person icon isn't intuitive and we should change it to a sign in button but it can't have the `md-raised` class.

monotkate:
Adding the words "Login" on my next PR as soon as I get tests written.  This should make it foolproofingly clear.  It also does not look like shit. :+1:

PR #205, opened 2015-12-16
Changed Team Interaction

monotkate:
Team:
- Now has consistent controls with backing
- Changed email button to use ng-click because of formatting issues with ng-href
- On removal of team, it alerts that the back will also be deleted and asks for confirmation
- There is no way to leave the team without deleting your back

Back:
- Added a dialog window that asks for confirmation when deleting your back (so as not to do it accidentally)

Tests:
- Checks for $mdDialog.show()
- Checks that backs delete correctly if user
- Checks that removal from team works correctly

Login
- Added "Login" to make it painfully obvious how you login

I think that's it.  I shouldn't keep adding onto this one.  Probably going to take a break for a bit to restructure my app for a potential January release.  One note for discussion: I think the email team button is kind of weird on the idea page.  I understand if on the team tab, but on the idea page I would expect it to email the idea author, not the whole team.

Fixes issue #183


src/ideas/ideasView/ideasView.js
```diff
@@ -593,6 +688,17 @@ angular.module('flintAndSteel')
...
+
+            // Open up an email to team members
+            ctrl.parseTeamEmail = function parseTeamEmail() {
+                var emailString = "mailto:";
+                $scope.idea.team.forEach(function(teamElement) {
+                    if (teamElement.member.mail !== 'undefined') {
+                        emailString += teamElement.member.mail + ';';
+                    }
+                });
+                $window.location = emailString;
```
Mctalian: Ah, clever!

Mctalian:
Approved pending the "Remove user from team when they delete their back" portion of the card.

monotkate:
@Mctalian I added it already.....

Mctalian:
:racehorse: that was fast! I'll check it out

monotkate:
@Mctalian Alright, got it.  Fixed clearing the back too, found a bug there after it's removed.  I missed that part and it worked but I didn't open the console error

Mctalian:
This looks good, thanks for fixing that. We'll have to figure out what to do with that "Owner" tag eventually. I see you are adding it if the author backs, which is good for now.

Approved :+1:

monotkate:
Yeah took me a bit to figure that one out since it wasn't in chip types. Awesome. Just need another approve now!

Mctalian:
@bcbrennecke

bcbrennecke:
I checked this over with Bob, looks great!  Approve!

PR #203, opened 2015-12-15
IE Compatibilities

Mctalian:
The following are now IE compatible (somehow they stopped being compatible at some point).
* About Page and closes #185
* Login Dialog RA-INT message
* Team Tab
* Edit Idea
* Add Idea

PR #206, opened 2015-12-17
Beef Up Testing!

Mctalian:
Ok, so this does a lot with our testing infrastructure.

### Test Warnings
If you can recall, there were test warnings that would pop up due to attempted calls to the server for our Server-Sent Event (SSE) listeners. These connect to the server with a "stay-alive" connection so the server can say "Hey! Here's new data" and we can act on it. When running these tests, however, we don't have a server up. So those warnings were popping up for "404" or "not found" errors.

What I did was wrapped the EventSource (SSE) stuff into a service called `sseSvc` in the `utilities` folder. So now those warnings have gone away!

### Code Coverage
I've been wanting to try Code Coverage to see how nice it is on web dev. Turns out, very. It actually inspired me to shoot for green (~80%, I think) on all of our files. And I *mostly* succeeded. I think @monotkate has some tests that would push us closer on some files. We'll see! Here's a look at what it looks like with this PR:

image

And this is what it looked like before:

image

I don't know if we'll develop any process around this, but I thought it'd be nice to have a visual representation of our code quality (at least somewhat).

#### Ok, so what do I need to do now?
Nothing! Well, run an `npm install` once this goes through to get a few dev dependencies, but once you have those, absolutely nothing. You can check out coverage results in the aptly named `coverage` folder, which is included in the `.gitignore` already.

### New Tests!

I wrote a lot of new tests for small files that didn't have any or for tests that were near green and just needed an adjustment or two to get there.

One important addition that I made was directive testing capability. We always had this option but testing directives are a bit tricky. I wrote some very basic tests for `emailLink`, `ideaItem`, `ideaItemBrowse`, etc. Basically enough to get "coverage" but because these are more complex in the HTML than the JS code, it doesn't do a whole lot. We could add some stuff to these later and it could be helpful in the future now that we have some things to reference.

### One last thing
This also has some of the commits from my IE PR that's open. So I'll need to rebase once one of those goes through.


Mctalian:
Oh, @YashdalfTheGray , @alanlgirard , @davethenipper 

Review is appreciated, though I'd prefer IE PR #203  goes in first since it's smaller. And also, probably #205 as well since I did all this, I'll handle the conflicts.

Mctalian:
I added what the current coverage (with the addition to my IE compatibility PR code FWIW) looks like for comparison.

davethenipper:
Holy shit. I love it

Mctalian:
I received some thumbs up on Skype while the internet was down @alanlgirard , @monotkate , and "MERGE IT!" from @YashdalfTheGray 

So I will do just that :shipit:

Issue #207, opened 2015-12-17
IE Add Backs Issue on logout


alanlgirard:
On Internet Explorer, if a user adds a back then logs out the back is not shown on the backs tab.  If the user then logs back in, they are able to add a back as if they didn't add one before.  Once they add the "2nd" back both backs show up.  This works properly on Chrome.

@monotkate @Mctalian can either of you take a look into this since you two have the most interaction with the backing?

Mctalian:
I have not been able to reproduce this.

davethenipper:
From discussion:

Refreshing seems to work: Some sort of weird caching issue?

Or IE isn't getting the updated data from the server in the database?

Daviddt17:
I think I solved the issue. Caching in Angular's $http is enabled in IE, so it was a matter of disabling it in the app config. I just pushed the change to my fork, but I do have an issue with not using dot notation.

I'll go ahead and make a PR. Here's the stack overf... I MEAN MY BRILLIANT SOLUTION TO THIS PROBLEM :)

http://stackoverflow.com/a/19771501

davethenipper:
HAHAHAHAHAHAHAHAHAH, amazing work!!!  I guess I need to go through that
phase 2 backlog sooner rather than later.

Mctalian:
Damn, that's awesome. Good work!

Test it out pretty thoroughly, I'm hoping this doesn't cause any adverse behavior anywhere else but I think it'll be pretty good. Damn IE. I wonder what else we'll run into...

Hardening phase will need to be completed on both Chrome and IE. Might as well go with Firefox too.

monotkate:
nice! :+1:

Daviddt17:
Closed with PR #211

PR #211, opened 2015-12-21
Disable IE angular $http chaching

Daviddt17:
This is the solution to #207.

Since some of the default fields include hyphens, how do I go about using dot notation since that is demanded by our style checker?

@YashdalfTheGray @Mctalian @davethenipper @alanlgirard @LongArmMcGee @bcbrennecke @monotkate

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -13,8 +13,8 @@ angular.module('flintAndSteel', [
     ]
 )
 .config([
-    '$urlRouterProvider', '$stateProvider', '$mdIconProvider', '$mdThemingProvider',
-    function($urlRouterProvider, $stateProvider, $mdIconProvider, $mdThemingProvider) {
+    '$urlRouterProvider', '$stateProvider', '$mdIconProvider', '$mdThemingProvider', '$httpProvider',
+    function($urlRouterProvider, $stateProvider, $mdIconProvider, $mdThemingProvider, $httpProvider) {
         "use strict";
 
         $stateProvider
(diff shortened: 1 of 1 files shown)


Mctalian:
Kind of hacky, but can you set a new var with the hyphenated string and use that var in the square bracket notation? Another option could be disabling that check for this block of code, you'd have to look up documentation for how to do that. I assume this is jscs not jshint.

Mctalian:
Both of those suggestions are guesses at best. So I could be way off here :)

Mctalian:
http://jscs.info/overview

Check out "Error Supression". That's one way. If it's Jshint, I think they have something similar.

I'm not sure if this falls under snake case because apparently we can ignore dot notation requirements for those but they seem to be underscored not dashed. (`snake_case_example`)

YashdalfTheGray:
I would go with the first suggestion. I'd like to not have rule disabling littered around the codebase.

src/app.js
```diff
@@ -99,6 +99,16 @@ angular.module('flintAndSteel', [
...
+
+        // Initialize get if not there
+        if (!$httpProvider.defaults.headers.get) {
+            $httpProvider.defaults.headers.get = {};    
+        }
+
+        // Disable IE ajax request caching
+        $httpProvider.defaults.headers.get['If-Modified-Since'] = 'Mon, 26 Jul 1997 05:00:00 GMT';
+        $httpProvider.defaults.headers.get['Cache-Control'] = 'no-cache';
+        $httpProvider.defaults.headers.get['Pragma'] = 'no-cache';
```
alanlgirard: To fix the jscs error this line should be $httpProvider.defaults.headers.get.Pragma = 'no-cache';
Jscs will complain because its only 1 word and should be using the dot notation instead of the bracket notation for vars that have hyphens.

Mctalian: A good call, I should have looked at the actual build results. Yup, that's an easy fix.

Daviddt17: Ahh that's what I get for being lazy.

Daviddt17:
I tested this solution on IE, Chrome, and Firefox. Seems like everything is in good order.

davethenipper:
Approved!

alanlgirard:
Approved

Daviddt17:
Merged!

Issue #166, opened 2015-11-24
Long Text in Comments, Backers, Updates eventually get truncated


Mctalian:
We should probably leave people with a text area for typing and text-wrapping for viewing a comment because people like to type a lot. For instance, if this text was on our site, I'm pretty certain, to the 99.9% confidence interval, that the text would be truncated when viewing it in the comments, backs, or updates view. I think we should allow our users the ability to type their little hearts out in the event that they have a lot to say, even if what they say doesn't seem to add any value. It is like the age old Zen saying on attachement: "The tighter you squeeze, the less you have." Ipso facto, we need to allow for longer comments, backs, and statuses without truncating them.

alanlgirard:
Maybe implement a "see more..." link at the end of long comments etc like Facebook does to allow the site to not become too cluttered, but I approve of allowing the user to not be limited.

Daviddt17:
We should also word wrap on input for updates and comments. As of now it's a single line.

Daviddt17:
Also, the press enter to submit comment feature will have to be deleted or done in some other way. The user needs a way to enter newlines.

Mctalian:
Good point. Maybe we can use "shift+enter" to post a comment. Or some other
modifier key.

Daviddt17:
"shift-enter" sounds perfect.

YashdalfTheGray:
Change the `<input>` tag inside the `<ng-input-container>` tag to a `<text-area>` tag to resolve this. Make sure to put some kind of a word limit on it too.

Mctalian:
@alanlgirard mentioned he was planning on working on this soon/next in the standup. Feel free to unassign yourself if I misunderstood.

alanlgirard:
Closed with PR #221

PR #221, opened 2015-12-22
IdeaPage Tab Overhaul/Overflow Control

alanlgirard:
Overhauled the idea page tabs to look a little better and to wrap properly.  The updates and comments now show as if they are dialogues. The backing tabs and team tabs were re-factored to take on the same kind of look.  

Update/Comments Page
image

Backers Tab
image

Team Tab
image

Overflow Areas
* All ideasView tabs
* Add/Edit idea description
* Add/Edit Back details
* Edit Team chips
* Leave Team button placement on small screens

As a bonus I fixed some of the form validation that was not working correctly before (back description) and made it so the login doesn't "bump" down anymore because the ng-messages were only showing on the touched state.

Comment/Approve away @Mctalian @monotkate @YashdalfTheGray @bcbrennecke


alanlgirard:
Will close issue #166

Mctalian:
Weren't we able to delete comments before? Or am I crazy?

alanlgirard:
You can still delete comments.  I got rid of the hover-over effect and now the delete comment is the x in the upper right of each comment/update

Mctalian:
Also, I think a little extra padding between the text box at the top of updates/comments and the first update/comment would help. Not too much, just a bit.

Overall, I like it. There are some things that feel a bit off but it's a major improvement over the old stuff. I'll try to put those things into words soon. Nice work!

monotkate:
- We need a way to highlight the controls at the top now.  Potentially their own card to box them in?
- I'm not a huge fan of the team tab restructure.  Potentially split it into the users like the other tabs instead of all of them in the same team area?  Also, I think we should split the email into single users then (although really I think this should only be an option if you're already on the team).  
- I don't like the X next to the name either, but if you restructure it a bit more like the comments, I think it could match them and look a lot better.
- More padding above for updates/comments just like @Mctalian said.  It's cutting it a but close.  Or, just add padding to the bottom of the entry. That would work too.

+ Overall a huge fan of the restructuring.  I like how the controls are embedded in each portion :100: It feels way more modernized this way.

YashdalfTheGray:
This...this...my dear lord this looks good! I love the embedded controls, totally fits in with Material Design as well!

src/assets/app.css
```diff
@@ -300,3 +300,82 @@ img.grayscale:hover {
...
+}
+
+.triangle-isosceles.left:after {
+  top:16px; /* controls vertical position */
+  left:-10px; /* value = - border-left-width - border-right-width */
+  bottom:auto;
+  border-width:10px 10px 10px 0;
+  border-color:transparent #e0e0e0;
+}
+
```
YashdalfTheGray: Check your spacing on the `.triangle-isosceles`, `.triangle-isosceles.left`, `.triangle-isosceles:after` and `.triangle-isosceles.left:after` classes. I think you've got 2 spaces.

YashdalfTheGray:
I think all the team members being in one thing makes sense since they're all related (since they are on a team together). Maybe see if you can get the 'x's to be aligned to the right? I think that would fit better with Material Design guidelines anyway. 

More padding could be nice, would probably look better, I'm not too opposed to the way it is right now though.

davethenipper:
-trashcan for deleting
-team members being separate

PR #222, opened 2015-12-23
REST API, promises, and services layer rewrites

YashdalfTheGray:
Alright, so there's a lot here but I'll put down as much information as I can and we can open discussions around the information. Go in and look around, play with it, pull it down and test things that I might have missed but please pull it down responsibly :smile:

Also, since this is a rather large PR, I'm going to require 4 approvals - @Mctalian @monotkate @davethenipper @alanlgirard 
Everyone should look at this though...@bcbrennecke, @Daviddt17, @LongArmMcGee

### REST API
The server code now has a fully RESTful API supporting CRUD operations as much as possible. What does this mean? This means that RokStarter has an API. What does that mean? If you want to create an Android app, you can! iOS? Yeah! Kindle Fire TV? If there are dev tools, you can do it (full disclosure, I *don't* believe in you). Fallout 4 Terminal?

Now you might be asking yourself, how do I get at this data? There is a handy dandy guide called [data-access.md](https://github.com/YashdalfTheGray/flintandsteel/blob/master-author/server/data-access.md) in the server folder which walks you through how to access certain kinds of data from the server. It lays out all the endpoints and all the operations that can be run on them. It even has examples that use Angular's `$http` service!

Another change is that now the lists of nested documents (i.e. `likes`, `backs`, etc.) now have ids too. So `idea.backs[2]._id` will return a mongo unique id and can then be used to run PATCHes against that back!

### Promises 
Promises are here and they're sticking around. So, a good chunk of the API route facing functions and methods have been converted to promises and some further down the callstack have been too! In the future, please stick to promises for creating new asynchronous functions and I will also be going through the rest of the server code and converting it all to promises. To create a promise in node (it's a little different than Angular and we use the `bluebird` npm-module):

```javascript
var Promise = require('bluebird');

function getIdea(id) {
    return new Promise(function(resolve, reject) {
        // use resolve or reject here just as you would
        // with Angular's deferred API.
    }
}

getIdea(myId).then(function(results) {
    // results here
}).catch(function(error) {
    // error here
});
```

Little different but still pretty familiar. There are also other real examples littered through the server code. Gone are the days of node style async callbacks. 

### Services Layer rewrites
To match the changed API, I also had to change the services (`userSvc` and `ideaSvc`) layer of the Angular app. First things first, I renamed `loginSvc` to `userSvc` because it does more than just login people now. 

Second, the services have been updated to use the new API while still mostly preserving how they were being used. The one change I do remember (I think it might be the only one) is editing a back now asks for `back._id` and not for `back.authorId`.  

Third, Angular is deprecating the `$http.get(url).success(successCb).error(errorCb)` syntax in favor of `$http.get(url).then(function(response) { // success }, function(response) { // error })` syntax. That should look familiar, `$http` now recommends promises and will soon stop using `.success()` and `.error()` so I updated our services to promises. 

**NOTE**: Since we're using promises in Angular, any time you are testing something that uses a promise in Jasmine, you have to call `scope.$digest()` or your tests will fail for no reason. `scope.$digest()` causes Angular to resolve/reject the promises in question. There are examples in the code for this too. 

### `gulp generate:data` rewrite
The `gulp generate:data` command now runs against the mongo shell in pure JavaScript. This decouples the server code from the client code from the gulp runner code and all that good stuff. As you can see from the commit comments, there has been some discussion around the safety of this and I'm of the view that the person running this is going to need to understand what the code is doing and why it failed. As developers, we should know what the commands that we create for ourselves do so when they break, we can fix them. 

It also makes sure that no matter how broken the server/client code gets, we can still get in and generate some test data to be able to get to the problem faster. 

### Known issues
@Mctalian I'm going to need your help on this, I've got most everything working except for the server side eventing stuff. Whenever the server tries to send a server side event, the server console shows this error (I shortened the paths a bit and formatted it):
```
Unhandled rejection Error: Can't set headers after they are sent.
    at ServerResponse.OutgoingMessage.setHeader (_http_outgoing.js:335:11)
    at ServerResponse.header (flintandsteel\node_modules\express\lib\response.js:718:10)
    at ServerResponse.contentType (flintandsteel\node_modules\express\lib\response.js:551:15)
    at ServerResponse.sendStatus (flintandsteel\node_modules\express\lib\response.js:339:8)
    at flintandsteel\server\routes.js:140:17
From previous event:
    at flintandsteel\server\routes.js:138:17
    at Layer.handle [as handle_request] (flintandsteel\node_modules\express\lib\router\layer.js:95:5)
    at next (flintandsteel\node_modules\express\lib\router\route.js:131:13)
    at Route.dispatch (flintandsteel\node_modules\express\lib\router\route.js:112:3)
    at Layer.handle [as handle_request] (flintandsteel\node_modules\express\lib\router\layer.js:95:5)
    at flintandsteel\node_modules\express\lib\router\index.js:277:22
    at param (flintandsteel\node_modules\express\lib\router\index.js:349:14)
    at param (flintandsteel\node_modules\express\lib\router\index.js:365:14)
    at Function.process_params (flintandsteel\node_modules\express\lib\router\index.js:410:3)
    at next (flintandsteel\node_modules\express\lib\router\index.js:271:10)
    at flintandsteel\node_modules\body-parser\lib\read.js:121:5
    at invokeCallback (flintandsteel\node_modules\raw-body\index.js:262:16)
    at done (flintandsteel\node_modules\raw-body\index.js:251:7)
    at IncomingMessage.onEnd (flintandsteel\node_modules\raw-body\index.js:308:7)
    at emitNone (events.js:67:13)
    at IncomingMessage.emit (events.js:166:7)
    at endReadableNT (_stream_readable.js:905:12)
    at doNTCallback2 (node.js:441:9)
```
Got anything off the top of your head on this? I tried to get the sse stuff to look like it did before I went in and changed up the API, the only changes I made to them were converting them to utilize promises (methods in the `ideas` modules changed to promises so these had to be changed as well) and change the end points to `/sse/ideas` and `/sse/ideas/:id` for headers and specific ideas respectively.


generateData.js
```diff
@@ -1,127 +1,172 @@
...
-                }
-            }
-        );
-    });
-};
+// create collections
+db.createCollection('users');
+db.createCollection('ideas');
+db.createCollection('comments');
+db.createCollection('events');
```
Mctalian: Are these synchronous operations? Are they guaranteed to finish before we start adding data to them?

Mctalian:
I'm trying to work on getting SSEs working again but can anyone confirm that unliking an idea gives the following console error on the server side? (**Note:** because SSEs aren't working, you'll have to login, like, refresh, login, unlike)

`Error: Argument passed in must be a single String of 12 bytes or a string of 24 hex characters`

server/db.js
```diff
@@ -89,23 +89,21 @@ module.exports = function(dbName, cb) {
...
+            db.collection(collection).findOne({_id: objId}, function(err, doc) {
+                if (err) {
+                    reject(err);
+                }
+                else if (doc) {
+                    resolve(doc);
+                }
+                else {
+                    var errNotFound = "NOT_FOUND";
+                    resolve(errNotFound);
```
Mctalian: Not sure if we should resolve `NOT_FOUND` here... the old behavior was to callback with a non-null error (first arg). This causes replaceIds to throw a shit fit when an idea gets deleted and someone tries to navigate to it. So that needs to be solved one way or another.

Mctalian:
Yeah, you need to adjust the client-side of unliking with these changes. We're passing in the old like object `{userId: "567abfeae15d79c5ce2f03c7"}` instead of the new one with the mongo-generated `_id` field. So somehow that's gotta get changed... A bit more complicated now, probably need to do some client-side `likes.filter` for the correct `like`. Might affect performance in the long run.

Mctalian:
I got SSEs working again, I will pre-pend all of those line comments with [SSE] so you know.

server/routes.js
```diff
@@ -35,16 +40,130 @@ module.exports = function(app) {
...
+            });
+        }
+    });
+
+    app.delete('/ideas/:id', function(req, res) {
+        ideas.delete(req.params.id).then(function() {
+            res.sendStatus(204);
+            return ideas.fetch();
+        }).then(function(headers) {
+            IdeasInstance.updateHeaders(headers);
```
Mctalian: [SSE] `IdeasInstance.updateHeaders` does not exist. This should be `IdeasInstance.newHeaders` or change the name of the appropriate `Ideas.prototype`. Make sure all the calls are consistent either way.

server/routes.js
```diff
@@ -35,16 +40,130 @@ module.exports = function(app) {
...
+
+        Promise.all(promises).then(function(results) {
+            res.status(200).json(results);
+            return Promise.all([
+                ideas.get(req.params.id),
+                ideas.fetch()
+            ]);
+        }).then(function(ideaResults) {
+            IdeasInstance.updateIdea(ideaResults[0]);
+            IdeasInstance.updateHeaders(ideaResults[1]);
```
Mctalian: [SSE] `IdeasInstance.updateHeaders` does not exist. This should be `IdeasInstance.newHeaders` or change the name of the appropriate `Ideas.prototype`. Make sure all the calls are consistent either way.

server/routes.js
```diff
@@ -35,16 +40,130 @@ module.exports = function(app) {
...
+        }).catch(function(error) {
+            console.log(error);
```
Mctalian: [SSE] Highly recommend putting some chalk on this. There was an error message here that helped with the `updateHeaders is not a function` diagnosis. I used `bgRed`

server/routes.js
```diff
@@ -81,268 +200,201 @@ module.exports = function(app) {
...
-                else {
-                    sse("newHeaders", headersData);
-                }
+        function updateIdea(idea) {
+            if (idea === null) {
+                sse("updateIdea_" + req.params.id, idea, req.params.id);
+                return;
+            }
+            replaceIds.idea(idea).then(function(ideaData) {
+                sse("updateIdea_" + req.params.id, ideaData, req.params.id);
```
Mctalian: [SSE] This needs `ideaData[0]` passed in, not the array of the resolves. I don't know in which order the Promise.All passes back it's resolves in the array, but the 0th element seems to be "the most correct".

Mctalian:
[SSE] I discovered another potential bug by inspection, can you make the change for me? Unfortunately I can't make the line comments because you didn't touch that area of the file.

Bottom of `server\ideas\ideas.js` in at and around `Ideas.prototype.updateIdea`, we can't track a single newestIdea, we need to have an array because if two people like two different ideas at the same time I think that an idea will be overwritten and that would be horrible.

Line 114 with this PRs current changeset: `var newestIdeas = null;` changed to `var newestIdeas = [];`

Line 118 with this PRs current changeset: Remove the line `newestIdea = idea`.

Line 128 with this PRs current changeset, in between the if-elseif-else block where we set `key` and the if block where we check if we're emitting updates, insert `newestIdea[key] = idea;

Line 134 with this PRs current changeset, within the if block where we check if we're emitting updates, inside the `setTimeout` callback, change `newestIdea` to `newestIdea[key]`

Mctalian:
In my testing here's what I've found:

* Removing a Like is not working properly. Getting an error: `Error: Argument passed in must be a single String of 12 bytes or a string of 24 hex characters` so it's something with setting a mongo ObjectId. From what I see, we're passing in an undefined _id.
* Updates aren't being deleted properly either. Same error as likes, but I didn't look into it as heavily.
* Comment deletion is working fine!
* Back Edits are working fine!
* Back deletion is not working, and it looks like a nastier error than Updates and Likes. It's calling out a MongoError with $set. Shouldn't we be using $pull since the back is in an array of backs?
* Team Edits are not working, they have the same $set error mentioned above.
* Team Removal is not working, they have the same $set error mentioned above.
* Editing an Idea works fine!
* Deleting an idea does not navigate users away from the idea. This is an SSE issue that will be addressed in a Line Comment shortly.

Outside of functionality:
* replaceIds needs to get beefed up. We've introduced a lot of cases where it can choke. For instance, if all the ideas get deleted it crashes the server. Similarly if someone navigates to a bogus or previously deleted idea, it will drop a lot of console errors and I don't think the client-side reacts right either. It should say "Sorry, that idea has been deleted" or something. I can't check into that right now, though, sorry.

server/routes.js
```diff
@@ -35,16 +40,130 @@ module.exports = function(app) {
...
+                }
+            });
+        }
+    });
+
+    app.delete('/ideas/:id', function(req, res) {
+        ideas.delete(req.params.id).then(function() {
+            res.sendStatus(204);
+            return ideas.fetch();
+        }).then(function(headers) {
```
Mctalian: [SSE] We need `IdeasInstance.updateIdea(null, req.params.id);` in this `then` because we need to notify all subscribing clients that the idea has been deleted and it forces them to leave that state, thus closing that SSE connection as well.

monotkate:
I just read this novel 👍🏻

I will look at this as soon as I have a chance, but since I'm currently using my family that's in town to overhaul my house for selling, hiding on a corner with my computer would be frowned upon. May get to it later tomorrow when we break for Christmas. Maybe Christmas itself.

monotkate:
Just starting to take a peep at this.  I'm not sure what @Mctalian is looking at, or if I did something wrong pulling it up.  If I add/edit a back, it crashes the server.  Since Bob didn't have problems with this, I'm guessing it's my end.  Any ideas?

davethenipper:
I've found the same things Bob has.  Here are a few others that maybe someone can confirm?
When I back an idea, the entire idea goes blank.  It returns and is correctly populated on refresh or switching to another idea, then moving back to the original idea.

Liking an idea also causes the idea to go blank.  Same solution.

Updating an idea: Same thing.

Forming a team: Same thing.

Mctalian:
This is all Server Side Event stuff. The [SSE] changes should fix the
things you listed.

Also @monotkate, your server crashes might also be related to the [SSE]
changes I proposed above. In fact I'm almost certain they are.

generateData.js
YashdalfTheGray: Yeah, they are.

YashdalfTheGray:
The team editing and removing stuff is incompatible with the new API. The way we update teams is that we commit the entire `$scope.idea.team` object to the database but that's not how the new API works. We will have to create a bug ticket after this goes in to work on it. I think we'll have to change the editing/deleting process.

As for right now, I've patched it so that it works but if two people change the team at the same time, it will have data conflicts.

YashdalfTheGray:
Ideally, we should use `PATCH /ideas/{id}` with `{ "operation": "append", "path": "team", "value": JSON.stringify({ memberId: someUserId }) }` to add a person to the team and `PATCH /ideas/{id}` with `{ "operation": "delete", "path": "team/" + teamMemberToDelete._id }` to delete a team member. The process we have right now to edit/delete teams is incompatible since we pass the entire team object around whenever we want to edit/delete teams.

YashdalfTheGray:
SSE for headers now works, SSE for the idea, I'll troubleshoot after I take a shower. @Mctalian, you're right, I think replaceIds is throwing a shitfit and won't pass things back. Other than that
* Removing likes works
* Removing updates works
* Removing backs works
* Look at my comments up higher for the team stuff
* Deleting the idea will hopefully start working after I get idea SSE working. Otherwise, I'll look at it after idea SSE.

YashdalfTheGray:
Got SSE working for a specific idea but deleting an idea still runs replaceIds on the deleted idea. I changed up how the `NOT_FOUND` message from `db.js` is handled too, it is now rejected instead of resolved and somehow, replaceIds is still running. 

@Mctalian, might need your help here

As far as I can see, this is the last thing for this PR.

server/ideas/ideas.js
```diff
@@ -181,27 +111,27 @@ module.exports = function(db) {
...
         }
-        if (typeof oldKey === "undefined") {
+
+        if (!oldKey) {
             key = idea._id;
+            newestIdeas[key] = idea;
         }
         else {
             key = oldKey;
+            newestIdeas[oldKey] = 'IDEA_NOT_FOUND';
```
Mctalian: I think this is why replaceIds is still running for idea deletions.

We're passing in idea which is null for a deleted idea. When we *used* to pass that out, the event emission would exit early and return the needed data. I will tag the location I mean with [HERE].

server/routes.js
```diff
@@ -81,268 +206,207 @@ module.exports = function(app) {
...
-                    console.error(chalk.bgRed(err));
-                }
-                else {
-                    sse("newHeaders", headersData);
-                }
+        function updateIdea(idea) {
+            if (idea === null) {
+                sse("updateIdea_" + req.params.id, idea, req.params.id);
+                return;
+            }
```
Mctalian: [HERE] The if statement above this comment is where we used to handle deleted ideas. So you'll need adjust either this line or the last line I commented on in ideas.js

Mctalian:
Where was my comment about NOT_FOUND in db.js? My line comments should handle the "deleting ideas still runs replaceIds" issue.

server/routes.js
YashdalfTheGray: I could change this to `if (idea === 'NOT_FOUND`)` and it would work, right? Pretty sure but I want a sanity check to see if I'm not missing your point.

YashdalfTheGray:
So I fixed what you said about `NOT_FOUND` in `db.js`. Look at commit 5435a550a510f1fedfd783cd0fc130f505789c10. You said resolving it there would be a bad idea and I agree now, I was trying something out back then which didn't pan out. 

I'm working on getting the idea delete SSE to work right now.

server/routes.js
Mctalian: That should work I think. That is what I meant.

YashdalfTheGray: Perfect!

Mctalian:
Got it, yeah I remembered almost everything but was still in need of a bit
of context. Looks good.

YashdalfTheGray:
Alright, deleting an idea works. That should be the end of that. Review again @Mctalian, @davethenipper, @alanlgirard, @monotkate, @bcbrennecke, @LongArmMcGee, @Daviddt17! Looking for 4 approvals.

bcbrennecke:
@YashdalfTheGray : Team tested, Team approved.  Merge away pleaseeee

PR #223, opened 2015-12-28
Homepage Redesign!

bcbrennecke:
Okay here's the start of the redesign.  Let's get this merged in and i'll continue working on changes to make it more user friendly.
The code should be a little simpler to read this time because all of the content is being pulled from the js.  This will make it uber easy to use once we have events fully functional.  This page is going to change some more once we have events!  All of the timeline events should be automatic depending on what events are in the db.  I want to have a couple tags showing the active events and that will sort the ideas depending on the event.
Things that I want to do but AFTER this pull request goes through
-Bob we need to talk about the ng-focus for the search bar again.  Right now I still just have the ideas displaying if there is text in the bar.
-Anchor tags for the page.  They would look great as a couple little words to the right of the 'Innovation Challenge' Text.
-Maybe some others that I forgot.  

Check it out, let me know what you think, but I think the page and the code is pretty clean this time that we can get it merged in and then changes can continue to be made.
@Mctalian @monotkate @davethenipper


bcbrennecke:
Stupid CI build

monotkate:
you're still failing build :(  I can't run it right now

bower.json
```diff
@@ -28,6 +28,8 @@
     "moment": "2.10.3",
     "event-source-polyfill": "0.0.6",
     "identicon": "*",
-    "angular-identicon": "0.1.17"
+    "angular-identicon": "0.1.17",
+    "angular-horizontal-timeline": "*",
```
alanlgirard: Doesn't look like you are actually using angular-horizontal-timeline?  If you are not using it, it should be removed from here.

karma.conf.js
```diff
@@ -15,7 +15,10 @@ module.exports = function(config) {
 
         // list of files / patterns to load in the browser
         files: [
+            
             'src/lib/angular/angular.js',
+            'src/lib/angular-timeline/dist/angular-timeline.js',
+            'src/lib/fmp-card-directive/fmpCardDirective.js',
```
alanlgirard: If you are using fmpCardDirective it should also be in bower.json so that it is installed on npm install/Travis CI builds.  If not it should be removed.  What are all of the dependencies you are adding and using (looks like you are only utilizing angular-timeline)?

src/assets/app.css
```diff
@@ -258,30 +258,6 @@ img.grayscale:hover {
...
-}
-
-.small-padding-bottom{
-    padding-bottom: 15px;
-}
-
-.back-box {
-    width: 600px;
-}
-
```
alanlgirard: These should not be removed.

bcbrennecke: Added these back in.  Must have gotten deleted somehow when I split the homepage css into a separate file.

karma.conf.js
bcbrennecke: Good catches with the dependencies.  I had added a couple things to try and there were some lingering unused parts that I never took back out.  They are removed now.

src/app.js
```diff
@@ -8,6 +8,7 @@ angular.module('flintAndSteel', [
         'ngAnimate',
         'ui.router',
         'ngMaterial',
+        'fmp-card',
```
alanlgirard: This should be removed too since you are not using it

src/homeView/homeView.js
```diff
@@ -3,13 +3,53 @@
...
             $scope.navToBrowse = function navToBrowse() {
                 $state.go('ideabrowse');
             };
+
+            $scope.showIdeas= false;
+
+            $scope.searchFocus= function searchFocus(){
+                console.log("focus");
+                $scope.showIdeas=true;
+            }
```
alanlgirard: Needs an end delimiter aka a semi-colon `;`   Make sure that you run gulp jshint, gulp jscs, and gulp test:client before submitting pull requests to catch issues like this.

bower.json
```diff
@@ -28,6 +28,7 @@
     "moment": "2.10.3",
     "event-source-polyfill": "0.0.6",
     "identicon": "*",
-    "angular-identicon": "0.1.17"
+    "angular-identicon": "0.1.17",
+    "angular-timeline": "~1.6.1"
```
alanlgirard: The ~ should be removed. For production we will want to make sure that we stay with a single version.  The ~ will allow multiple versions to be installed which could in-turn break stuff.

bcbrennecke: Done.

src/app.js
bcbrennecke: Done.

src/homeView/homeView.js
bcbrennecke: Done.  Never used jshint, that's kinda cool

src/homeView/helpCard/helpCard.tpl.html
```diff
@@ -0,0 +1,13 @@
+<div id="card" class="containerCard">
+        <div>
+                    <div class="box-front" >
```
YashdalfTheGray: I think you have tabbing issues here.

src/homeView/homeView.css
```diff
@@ -0,0 +1,473 @@
+.homeview-Image {
+ height:100%;
+   width:100%;
```
YashdalfTheGray: Spacing issues in all the styles for the `.homeview-image` class as well as CSS classes are named in `snake-case` and not in `camelCase`. It should be `.homeview-image` instead of `.homeview-Image`.

src/homeView/homeView.tpl.html
```diff
@@ -1,24 +1,65 @@
...
+    <div layout="row" flex layout-wrap layout-align="center center" ng-hide="searchText">
+        <md-item ng-repeat="faq in faqCards" class="homeview-padding-faq-card">
+            <help-card  faq="faq"/>
+        </md-item>
+    </div>
+    <!--End Help Cards-->
+</div>
+<!-- Begin Idea Cards -->
+<div layout="row" layout-align="center center" ng-show="searchText">
+        <div flex="10"></div>
```
YashdalfTheGray: Tabbing issues here too.

YashdalfTheGray:
Couple of things I wanted to mention:
Some of the CSS that you have wierd formatting. CSS Classes should have one space between the class name and the curly brace and the curly brace should be on the same line. Every style inside it should be indented 4 spaces and the ending brace should be on a line by itself. See the example class below. There are a few classes in your CSS that don't follow this pattern. 
```css
.my-class {
    padding: 10px;
    color: #1e367d;
}
```

Can you write tests for the new functionality that you added to `homeView.js`?

And with Bob's recent PR, we can now test directives, your directives aren't that complex but we should still have basic tests for them. Can you include those too?

Also, can you post an image (maybe a .gif) for all the content that you added and the CSS animations just so your PR is fully documented?

bcbrennecke:
@YashdalfTheGray: fixed the css classes.  
I'm not sure there really is much to test, but I will need help writing those this week from either you or @Mctalian . 
You just want a picture of the homepage?

PR #229, opened 2016-01-11
Event Tags: Part 1 of 3

Mctalian:
This is the first of 3 deliveries around event tags. This one is just being able to interact with existing event tags. They were always present on the generated ideas, we just had no way of seeing them before. Isn't that neat :stuck_out_tongue_closed_eyes: 

The next delivery will be the ability to choose an event when you create/edit your idea.

The last delivery (if we think it's a good idea) is to make a page for each event that lists the data associated with that event: Start and End Dates, Location, and all of the ideas tagged under the event. Someone else could take that one if interested, I don't think it is blocked by anything, but I could be wrong. Might require a bit of backend work, but I think @YashdalfTheGray took care of everything that would be needed there.

The idea portion of the ideaView has been revamped. It also includes the name of the event that the idea belongs to.

### ideasView Revamp with event tags
ideasvieweventtag

Searching for ideas by event tag name is also now possible.

### Searching
searchingbyevent


Mctalian:
Need reviewers: @monotkate , @alanlgirard , @LongArmMcGee 

Since we all seemed to be interested in reworking ideasView layout, I'd like your opinions on my proposed changes.

monotkate:
I like where you moved the search tags.  A+

I don't think it's something you did, but do we really want it to wrap in the middle of a word?  Was that @alanlgirard ?

It looks kind of.... boring right now.  As soon as I get some time (hopefully this week) I can fix my damn tests and add the backing tags wanted.  That should at least add some color?

Mctalian:
Without some sort of image associated with the idea, it will look boring
unfortunately. I tried to get inspiration from KickStarter but they rely
heavily on a video or image

monotkate:
I feel like until we get that, we could spruce it up somehow.  Add a little color at least maybe?  Idk, for now we can stick to this and I can mess with it later too.  After my house is listed (in about a week) I should have all that free time back since I won't be able to touch anything in it....

monotkate:
Why is this part 1 of 3?

Mctalian:
It explains it at the top of the PR a little bit ;)

I also agree with you that roles requested or whatever we're calling it
should add some much needed color. Aside from that we'll think on it more.

monotkate:
I are smart.  Apparently I got interrupted enough this morning that I didn't realize I didn't read it.

I would say it looks good outside of fixing the wrapping of text.  @alanlgirard PR shows that it did not cut words in half on the ideas view, so fix that :smile:

Mctalian:
image

Mctalian:
BTW @monotkate that tag that I circled in the original picture was the *event* tag, not the search tags that already existed. At the moment, those still exist at the bottom under the description.

monotkate:
ooooo i didn't notice that.  Sweeet.  I like it, and you fixed the word wrap.  Who knew!  Ship it :+1:

Mctalian:
Anyone else want to look at this? Only got one review/approval...

@LongArmMcGee, @alanlgirard, @YashdalfTheGray need another approval so I can merge this and move onto the second step.

YashdalfTheGray:
Approved!

Mctalian:
I broke the build, I'm fixing it

monotkate:
⭐️

PR #238, opened 2016-01-15
Fixing the broken build

Mctalian:


Mctalian:
So Travis didn't catch it before because changes were made after my last changes, so Travis had no reason to run. It's triggered off of a commit to a branch that it's watching. I should have rebased first, but I saw the check mark and the approval and went all gungho.

In the future, that can be some useful knowledge.

Mctalian:
Merged to fix the build.

monotkate:
Good thing to know 👍🏻

PR #241, opened 2016-01-16
Event Tags: Part 1.5 of 3

Mctalian:
## Acceptance Criteria
- [ ] Should be able to select an event when editing an idea
- [x] ~~Should be able to select an event when creating an idea~~
- [x] ~~Event tag should be visible when viewing an idea if the idea is assigned to an event~~

## Coverage
image

## Demo
add_event

## Notes about the Feature Portion of the PR
I finished having the user able to select an event when creating a new idea. I will get to the idea editing portion at a later date.

There are also some additional checks we can put in place. For instance, no one should be able to tag their idea with an event id that doesn't exist in the database (if the id is empty `""` that's fine though). You wouldn't be able to do this from the UI, but there may be some sneaky ways. What is the risk? The database wouldn't be able to find the specified eventId and would return `EVENT_NOT_FOUND` I think. That may make `replaceIds.js` choke and the idea might not display properly. Or nothing could happen, I'm not sure.

## Notes about general cleanup in this PR
* I made some test specs DRYer by taking repeated code and putting it in a beforeEach
* I added `rootScope.$digest()` to tests that rely on promises being resolved or rejected. Apparently that's needed when testing promises.
* I simplified the logic for finding the users ideas for accountView (currently unused). It now uses `.filter()` instead of looping through and checking manually.

### DIFF
--- a/src/ideas/addIdeaView/addIdeaView.js
+++ b/src/ideas/addIdeaView/addIdeaView.js
@@ -4,8 +4,8 @@
 angular.module('flintAndSteel')
 .controller('AddIdeaViewCtrl',
     [
-        '$scope', '$state', 'toastSvc', 'ideaSvc', 'userSvc',
-        function($scope, $state, toastSvc, ideaSvc, userSvc) {
+        '$scope', '$state', 'toastSvc', 'ideaSvc', 'userSvc', 'eventSvc',
+        function($scope, $state, toastSvc, ideaSvc, userSvc, eventSvc) {
             "use strict";
 
             if (!userSvc.isUserLoggedIn()) {
@@ -15,8 +15,17 @@ angular.module('flintAndSteel')
 
             $scope.idea = {};
             $scope.idea.tags = [];
+            $scope.idea.eventId = "";
             $scope.tagInput = "";
 
+            var nullEvent = {
+                _id: "",
+                name: "No Event"
+            };
+
+            ///////////////////
+            // TAG FUNCTIONS //
+            ///////////////////
             $scope.doesTagExist = function doesTagExist(tag) {
                 if ($scope.idea.tags.indexOf(tag) === -1) {
                     return false;
@@ -48,9 +57,26 @@ angular.module('flintAndSteel')
                 }
             };
 
+            /////////////////////
+            // EVENT FUNCTIONS //
+            /////////////////////
+
+            $scope.loadEvents = function() {
+                eventSvc.getEvents().then(function getEventsSuccess(response) {
+                    $scope.events = [nullEvent];
+                    $scope.events = $scope.events.concat(response.data);
+                }, function getEventsError(response) {
+                    $scope.events = [];
+                    console.log(response);
+                });
+            };
+
+            ////////////////////
+            // IDEA FUNCTIONS //
+            ////////////////////
+
             $scope.addNewIdea = function addNewIdea(ideaToAdd) {
                 ideaToAdd.authorId = userSvc.getProperty('_id');
-                ideaToAdd.eventId = "";
                 ideaToAdd.rolesreq = [];
                 ideaSvc.postIdea($scope.idea).then(function postIdeaSuccess(response) {
                     if (angular.isDefined(response.data.status) && response.data.status === 'Created') {
--- a/src/ideas/addIdeaView/addIdeaView.js
+++ b/src/ideas/addIdeaView/addIdeaView.js
@@ -63,8 +63,7 @@ angular.module('flintAndSteel')
 
             $scope.loadEvents = function() {
                 eventSvc.getEvents().then(function getEventsSuccess(response) {
-                    $scope.events = [nullEvent];
-                    $scope.events = $scope.events.concat(response.data);
+                    $scope.events = [nullEvent].concat(response.data);
                 }, function getEventsError(response) {
                     $scope.events = [];
                     console.log(response);
--- a/src/ideas/addIdeaView/addIdeaView.js
+++ b/src/ideas/addIdeaView/addIdeaView.js
@@ -28,7 +28,7 @@ angular.module('flintAndSteel')
                 var reNonAlpha = /[.,-\/#!$%\^&\*;:{}=\-_`~()<>\'\"@\[\]\|\\\?]/g;
                 tag = tag.replace(reNonAlpha, " ");
                 tag = _.capitalize(_.camelCase(tag));
-                if ($scope.idea.tags.length !== 5 && !$scope.doesTagExist(tag) && tag !== '') {                    
+                if ($scope.idea.tags.length !== 5 && !$scope.doesTagExist(tag) && tag !== '') {
                     $scope.idea.tags.push(tag);
                 }
             };
@@ -43,7 +43,9 @@ angular.module('flintAndSteel')
 
             $scope.removeTag = function removeTag(tag) {
                 var index = $scope.idea.tags.indexOf(tag);
-                $scope.idea.tags.splice(index, 1);
+                if (index >= 0) {
+                    $scope.idea.tags.splice(index, 1);
+                }
             };
 
             $scope.addNewIdea = function addNewIdea(ideaToAdd) {
--- a/src/ideas/addIdeaView/addIdeaView.spec.js
+++ b/src/ideas/addIdeaView/addIdeaView.spec.js
@@ -9,14 +9,14 @@
 describe('AddIdeaViewCtrl', function() {
     "use strict";
 
-    var scope, rootScope, $q, ctrl, toastSvc, $state, ideaSvcMock, userSvcMock;
+    var scope, rootScope, $q, ctrl, toastSvc, $state, ideaSvcMock, userSvcMock, eventSvcMock;
 
     beforeEach(module('flintAndSteel'));
     beforeEach(module('ui.router'));
     // needed because $state takes us to home by default
     beforeEach(module('homeView/homeView.tpl.html'));
 
-    beforeEach(inject(function($rootScope, _$q_, $controller, _toastSvc_, _$state_, _ideaSvcMock_, _userSvcMock_) {
+    beforeEach(inject(function($rootScope, _$q_, $controller, _toastSvc_, _$state_, _ideaSvcMock_, _userSvcMock_, _eventSvcMock_) {
         rootScope = $rootScope;
         scope = $rootScope.$new();
         $q = _$q_;
@@ -24,6 +24,7 @@ describe('AddIdeaViewCtrl', function() {
         $state = _$state_;
         ideaSvcMock = _ideaSvcMock_;
         userSvcMock = _userSvcMock_;
+        eventSvcMock = _eventSvcMock_;
 
         spyOn($state, 'go');
 
@@ -32,14 +33,16 @@ describe('AddIdeaViewCtrl', function() {
             $state: $state,
             toastSvc: toastSvc,
             ideaSvc: ideaSvcMock,
-            userSvc: userSvcMock
+            userSvc: userSvcMock,
+            eventSvc: eventSvcMock
         });
 
         scope.idea = {
             title: 'Test Title',
+            eventId: 0,
             description: 'This is a test idea.',
             tags: ['TestTag1', 'TestTag2']
-        }
+        };
     }));
 
     afterEach(function() {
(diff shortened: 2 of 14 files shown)


Mctalian:
**Reviewers (2/4):** @monotkate (I increased coverage and touched some tests), @YashdalfTheGray (general), @alanlgirard (I touched accountView), @Daviddt17 (I made minor changes to tagging because there was a chance that someone could delete a tag that didn't exist and it would remove one from the end. Not even sure how they'd do that, but I put in a check to prevent it)

src/ideas/ideasView/ideasView.js
```diff
@@ -72,9 +72,6 @@ angular.module('flintAndSteel')
                     }
                     else {
                         $scope.idea = response.data;
-                        if (typeof $scope.idea.team === "undefined")	{
-                            $scope.idea.team = [];
-                        }
```
Mctalian: @davethenipper , @YashdalfTheGray I didn't think this was needed anymore. It was a temporary thing when we put it in until backend was hashed out. Just keep me honest here and make sure I'm not an idiot.

YashdalfTheGray:
General comment, don't use `rootScope.$digest()` unless you don't have a scope declared in the tests. If there is a `scope` available, run `scope.$digest()` instead of `rootScope.$digest()`. The reason being that running a `$digest()` on `rootScope` runs a digest cycle on everything in the app instead of just the view that we are testing. 

At this point, I don't think we have too many components using the `controllerAs` syntax so you should have a `scope` available in most test specs.

src/ideas/ideasView/ideasView.js
YashdalfTheGray: You know, when I was going through the `ideasView` page updating it for the new backend, I wondered what this was for and I couldn't see it. I thought to myself that I was going to bring it up but never did. With the new schema, I don't think we would ever hit that so yeah, remove it I would say.

@davethenipper, what are your thoughts?

src/ideas/addIdeaView/addIdeaView.spec.js
```diff
@@ -9,152 +9,209 @@
...
             tags: ['TestTag1', 'TestTag2']
         };
-        scope.addNewIdea(idea);
+    }));
 
-        expect(ideaSvcMock.postIdea).toHaveBeenCalled();
-        expect(idea.eventId).toBe("");
-        expect(idea.rolesreq.length).toBe(0);
+    afterEach(function() {
+        rootScope.$digest();
```
YashdalfTheGray: Make this `scope.$digest()`, does the same thing but only on the controller we are testing.

src/ideas/addIdeaView/addIdeaView.js
```diff
@@ -43,12 +52,31 @@ angular.module('flintAndSteel')
...
             };
 
+            /////////////////////
+            // EVENT FUNCTIONS //
+            /////////////////////
+
+            $scope.loadEvents = function() {
+                eventSvc.getEvents().then(function getEventsSuccess(response) {
+                    $scope.events = [nullEvent];
+                    $scope.events = $scope.events.concat(response.data);
```
YashdalfTheGray: You could switch these lines to `$scope.events = [nullEvent].concat(response.data)`. I know, kinda nitpicky but just had a thought. I don't want to force your style or anything.

src/ideas/addIdeaView/addIdeaView.spec.js
```diff
@@ -9,152 +9,209 @@
...
-            title: 'Test Title',
-            authorId: 3,
-            description: 'This is a test idea.',
-            tags: ['TestTag1', 'TestTag2']
-        };
-        scope.addTag('testTag3');
+    describe('scope.loadEvents', function() {
+        it('should populate the events', function() {
+            scope.loadEvents();
+            rootScope.$digest();
```
YashdalfTheGray: Here too, change `$rootScope.$digest()` to `scope.$digest()`.

src/ideas/addIdeaView/addIdeaView.spec.js
```diff
@@ -9,152 +9,209 @@
...
-            tags: ['TestTag1', 'TestTag2']
-        };
-        scope.removeTag('TestTag2');
+        it('should append "none" to the list of events', function() {
+            var numEvents;
+            eventSvcMock.getEvents().then(function(response) {
+                numEvents = response.data.length;
+            });
+            scope.loadEvents();
+            rootScope.$digest();
```
YashdalfTheGray: Change to `scope.$digest()` here as well.

src/ideas/addIdeaView/addIdeaView.spec.js
```diff
@@ -9,152 +9,209 @@
...
+        it('should set events to an empty array if there was an error', function() {
+            spyOn(eventSvcMock, 'getEvents').and.callFake(function() {
+                return $q.reject('FAIL');
+            });
+            spyOn(console, 'log').and.callFake(function() {});
 
-        expect(scope.idea.tags.length).not.toBe(3);
-        expect(scope.idea.tags.length).toBe(2);
+            scope.loadEvents();
+            rootScope.$digest();
```
YashdalfTheGray: Change to `scope.$digest()`.

src/ideas/addIdeaView/addIdeaView.spec.js
```diff
@@ -9,152 +9,209 @@
...
-        expect(scope.idea.tags.length).toBe(5);
+        it('should log an error if the idea cannot be created', function() {
+            spyOn(ideaSvcMock, 'postIdea').and.callFake(function() {
+                return $q.reject('FAIL');
+            });
+            spyOn(console, 'log').and.callFake(function() {});
+
+            scope.addNewIdea(scope.idea);
+
+            rootScope.$digest();
```
YashdalfTheGray: Change to `scope.$digest()` here as well.

src/ideas/ideasView/ideasView.spec.js
```diff
@@ -159,13 +159,16 @@ describe('IdeasViewCtrl', function() {
             updatesLength = scope.idea.updates.length;
         });
 
+        afterEach(function() {
+            scope.$digest();
```
YashdalfTheGray: :+1:

src/users/accountView/accountView.js
```diff
@@ -21,14 +21,16 @@ angular.module('flintAndSteel')
...
                 // Find all User Ideas
-                for (var i = 0; i < response.data.length; i++) {  // was let
-                    if (userSvc.isUserLoggedIn() && userSvc.getProperty('_id') === response.data[i].authorId) {
-                        $scope.userIdeas.push(response.data[i]);
-                    }
+                if (userSvc.isUserLoggedIn()) {
+                    var userId = userSvc.getProperty('_id');
+                    $scope.userIdeas = response.data.filter(function(idea) {
+                        return userId === idea.authorId;
+                    });
```
YashdalfTheGray: You can use the `GET /ideas/search` endpoint for this instead of doing this client side. Check out [searching](https://github.com/YashdalfTheGray/flintandsteel/blob/master/server/data-access.md#searching) in `server/data-access.md`.

src/users/accountView/accountView.spec.js
```diff
@@ -36,6 +39,10 @@ describe('AccountViewCtrl', function() {
             });
         }));
 
+        afterEach(function() {
+            rootScope.$digest();
```
YashdalfTheGray: Change to `scope.$digest()` here.

src/users/accountView/accountView.spec.js
```diff
@@ -65,6 +73,8 @@ describe('AccountViewCtrl', function() {
                 userSvc: userSvcMock,
                 ideaSvc: ideaSvcMock
             });
+
+            rootScope.$digest();
```
YashdalfTheGray: Change to `scope.$digest()` here too.

YashdalfTheGray:
Side note, this is a good sized PR. Perfect for looking at in about 15 to 20 minutes, I wasn't burned out on staring at code, it stays on point (doesn't touch files unrelated to the theme) and you've probably minimized on merge conflicts as well. 

Double, nay, tripple kudos, @Mctalian!

src/ideas/addIdeaView/addIdeaView.js
Mctalian: No that's a good call. Product of incremental work. I didn't have a null event in my early version so I overlooked this obvious improvement.

src/users/accountView/accountView.js
Mctalian: We can look into that later, but that's a good point.

monotkate:
I like the test improvements :+1: 

What happens with the interface if there are like 20 open events?  Wondering if this is the best way to structure the front end portion for ease of usability

Mctalian:
I was already thinking about that as well. It might also be necessary that we tag an idea with up to 3 events in case the idea is in one innovation challenge or whatever and carries over into the next event or two. For instance if it's an idea in a generic challenge, that ends, then it applies to a security-focused event, we probably don't want them to remove the old tag so we have some sort of archive of recent events.

With that being said, I'm thinking a modal with check-boxes similar to backing, but that's future.

For now I think this is OK, but I'm glad there's more than one reason that this interface likely isn't long run.

That reminds me, I need to take the "end date" into account when listing those events. Next PR, which is nearly ready to go after this one is approved.

monotkate:
I like dialog, but if it's a dialog with checkboxes you might end up with massive amounts.  Especially if ultimately we want to push this as part of IPs and they may get tagged from those potentially?

Mctalian:
We can discuss this and change as needed. I can't think of a better way
right now. I'm going to leave it like this for now unless there's a better
idea.

Could get fancy and make a big dialog with a list of events as cards and a
search bar at the top, clicking an event "selects" it and changes color or
something and you can select up to 3.

If David T does part 3 of Event stuff, he can wrap it in a directive and we
can just use that (maybe with slight modifications) and it wouldn't be
hard. But I'm going to pull the "outside of PRs scope" card on that.

monotkate:
Pushing this issue as outside of PR scope makes me a bit concerned.  I think we're doing this a lot lately as an excuse to get things done instead of properly thinking out how best to carry it out, which creates a lot of rework down the road as well as opening up issues.  

I don't think we need to get fancy now even, but I think that we may need to add a search much like we were originally doing backing tags instead of just a drop down due to the number of events that would rack up in theory at least.

Mctalian:
It's not an excuse to get things done, it's just that we would be trying to solve a problem that hasn't and might not ever occur. Is it likely? That's arguable. Right now, we can guarantee at most 2 concurrent events. That's all we're slated for. A fix for this problem may not actually be better for the future. The solution for this eventual problem might not even be visible to us yet. But I do foresee the need to change event tagging prior to us running into the problem you're stating now. So, personally, I'm not going to optimize twice. It works now, we'll make it better when we need to.

YashdalfTheGray:
I think, even with a large number of events, a dropdown is the best way of doing this. The other choices we have create dialogs with a large number of items in them and radio buttons in front of said items. That's a lot of space that we could use more effectively. We might have to have a look-ahead search of some sort but I don't think that it will be required as of now. 

On the other point, I 100% agree with @monotkate that "out of scope" can be (and is) used as a cheap way to defer work and there are limitations to make it work then make it better, I don't think that it's happening in this PR. I could go into this and comment on our process here too but I'll just say this - Since this is pre-MVP, our commits and PRs are rather large, this problem being out of scope comes up a lot. Once we stabilize a little bit and actually do kanban, we would write cards in a way that the scope of the card will be well defined. If you're working on that card, you're not supposed to do any future work, you just implement what the card is asking you as best you can. That's how our cards should work ideally but our codebase and our plan is kind of in flux right now. 

Hopefully, this (larger) issue will resolve itself but if it doesn't, then we'll need to take a look at our process and see what else we can do to make it better.

monotkate:
@Mctalian I'll accept that.  It was nothing personal, just irritation at some of the PRs going through hastily.  I'm guilty of it myself and I goofed things up some with trying to push the test PR through quicklyin conjunction with @YashdalfTheGray 's code.  At the time I was frustrated in general.  I apologize.

Mctalian:
All good. I was also a little irked from testing the new directive I added. So my response may have seemed a bit cold. Sorry about that!

Mctalian:
Can I get one more approval?

YashdalfTheGray:
I didn't approve because there was one check box still empty but yeah, your
changes so far are good. Approved.

Yash Kulshrestha
yash.kulshrestha@gmail.com
(614)563-1163

Mctalian:
Oh whoops, I counted one of your comments above as an approval. I was hoping to take care of that check box in a separate PR because the changes build off of this one, but things changed around a bit. It's kind of hard to explain.

YashdalfTheGray:
Alright no worries. Besides the things that I commented on, it really was
an approval but I was just making sure.

Yash Kulshrestha
yash.kulshrestha@gmail.com
(614)563-1163

alanlgirard:
Looks good.

Approved :+1:

PR #249, opened 2016-01-20
Fixed DialogBackCtrl

monotkate:
I'm pretty sure I'm the one that broke this in the merge confusion between @YashdalfTheGray and I, but I caught it doing a rebase today.  Small fix,  but without it adding and editing a back won't work.

@YashdalfTheGray @alanlgirard @Mctalian


monotkate:
Wait, I think more than this is broken, at least on mine.  Fixing now + writing tests for it, I'll update this.

monotkate:
ok, writing tests later, this should fix it now, and if it fails the build I have no clue why.....

Mctalian:
It works now right? If so, approved.

monotkate:
yes it does.  Which is a good thing

alanlgirard:
Approved

Issue #248, opened 2016-01-19
As a user, I need to be notified when I am tagged in a conversation by another user so that I can respond


davethenipper:
Acceptance Criteria:
- [ ] To Come

PR #250, opened 2016-01-21
endpoints are now mounted under "/api/v1"

YashdalfTheGray:
Not much to say, changed our REST API mount point to `api/v1` instead of `/ideas` or `/users`, etc.

Reviewers (or mostly just look at it): @bcbrennecke, @monotkate, @alanlgirard

### DIFF
--- a/server/routes.js
+++ b/server/routes.js
@@ -40,7 +40,7 @@ module.exports = function(app, db) {
         };
     }
 
-    app.get('/ideas', function(req, res) {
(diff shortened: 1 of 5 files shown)


Mctalian:
I trust you were diligent and got them all. I'll have to change #248 for events. That's on me though.

Approve

alanlgirard:
Gotta love find and replace. 

 Approved

Issue #235, opened 2016-01-12
Comments and Updates able to be submitted blank


alanlgirard:
Need to add more error checking to handle the pristine/dirty states of all form fields to ensure that we only submit valid entries

YashdalfTheGray:
Is this a validation issue or do you think the server can take empty fields?

YashdalfTheGray:
Actually, I'm pretty sure that the server can take empty fields but if  we have some validation, we should be good for now.

alanlgirard:
It's just a validation check. We are limiting characters but since the
field is not required there needs to be a few other validation checks
because the form doesn't have a invalid state until the user enters text.

LongArmMcGee:
Meeting Decision: Make the field required instead.

PR #252, opened 2016-01-21
Form Validation Fix

alanlgirard:
- Fixed the login dialogue so the required error message only show after the fields have been touched and there is no weird bump down now
- Updates/Comments have the submit button disabled now if the field is empty.  I used this method over just making them required because I didn't like the fields being highlighted red after a user clicked on them.

Closes issue #235 

Very quick review @YashdalfTheGray @Mctalian @monotkate

### DIFF
--- a/src/ideas/ideasView/ideasView.tpl.html
+++ b/src/ideas/ideasView/ideasView.tpl.html
@@ -229,7 +229,7 @@
                                         <div ng-message="maxlength">Your comment is too long Must be under 500 characters</div>
                                     </div>
                                 </md-input-container>
-                                <md-button class="md-icon-button" type="submit" aria-label="Submit" ng-disabled="newUpdateForm.$invalid">
+                                <md-button class="md-icon-button" type="submit" aria-label="Submit" ng-disabled="newUpdateForm.$invalid || !ideasView.newUpdate">
                                     <i class="material-icons icon-button-adjust">&#xE876;</i>
                                 </md-button>
                             </form>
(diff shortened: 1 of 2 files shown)


src/ideas/ideasView/ideasView.tpl.html
```diff
@@ -229,7 +229,7 @@ <h1 class="no-margins" layout="row">{{idea.title}}</h1>
                                         <div ng-message="maxlength">Your comment is too long Must be under 500 characters</div>
                                     </div>
                                 </md-input-container>
-                                <md-button class="md-icon-button" type="submit" aria-label="Submit" ng-disabled="newUpdateForm.$invalid">
+                                <md-button class="md-icon-button" type="submit" aria-label="Submit" ng-disabled="newUpdateForm.$invalid || !ideasView.newUpdate">
```
Mctalian: Do you need these conditions with "required" as well?

Mctalian:
Approved pending my question being answered

src/ideas/ideasView/ideasView.tpl.html
alanlgirard: The form `.$invalid` conditional would still be required if "required" would be used for the form.

Mctalian: What about !ideasView.newUpdate and newComment? That's what I meant

alanlgirard: No those would not be required.  I only did it this way to prevent the angry wrath of the red input form once a user clicked off of it.

Mctalian: Ah ok, all good then.

monotkate:
Approved

PR #257, opened 2016-01-26
Event Tags: Part 2 of 3

Mctalian:
## Acceptance Criteria
- [x] ~~Should be able to select an event when editing an idea~~
- [x] ~~Should be able to select an event when creating an idea~~
- [x] ~~Event tag should be visible when viewing an idea if the idea is assigned to an event~~

## Coverage
image

## Demo
ideainput

## Feature
I moved all of the add new idea and edit existing idea form stuff into a single directive ideaInput. This made event tagging possible on both fronts. This combined a lot of duplicate code and tests. So a lot of the changes you'll see are deletions from where the old code lived.

Minor one line changes are because I updated the sseSvcMock to facilitate some better tests. For instance, it helped me improve ideasView coverage from <80% to >85% after I pulled all the well-covered, yet duplicated code out.

## Reviewers
@YashdalfTheGray (I tested a directive with a controller)
@alanlgirard (General)
@monotkate (I tested a directive with a controller)
@Daviddt17 (I consolidated most of the tagging stuff)


src/ideas/ideaInput/ideaInput.js
```diff
@@ -0,0 +1,23 @@
...
+                restrict: 'E',
+                scope: {
+                    idea: '=',
+                    submitFn: '&',
+                    submitBtnText: '@',
+                    cancelFn: '&',
+                    cancelBtnText: '@'
+                },
+                templateUrl: 'ideas/ideaInput/ideaInput.tpl.html',
+                controller: 'ideaInputView'
```
YashdalfTheGray: `ideainputView` isn't a view technically, it's a directive. So I would name it `IdeaInputCtrl` since it's also a constructor.

src/ideas/ideaInput/ideaInputView.js
```diff
@@ -0,0 +1,67 @@
+/* global angular */
+/* global _ */
+
+angular.module('flintAndSteel')
+.controller('ideaInputView',
```
YashdalfTheGray: I would name this file `IdeaInputCtrl` since it's a controller and this controller doesn't attach to a view; it attaches to a directive.

src/ideas/ideaInput/ideaInputView.spec.js
```diff
@@ -0,0 +1,163 @@
+/* global describe */
+/* global module */
+/* global beforeEach */
+/* global inject */
+/* global it */
+/* global expect */
+/* global spyOn */
+
+describe('ideaInputView', function() {
```
YashdalfTheGray: Same here, `IdeaInputCtrl`.

YashdalfTheGray:
The code and the tests look fine. This is a rather large PR but that's understandable since a lot of it is moving code around.

Mctalian:
Yeah it was kind of inevitable since I took two very large portions of copied code and consolidated them into one. I netted negative lines, so I deleted more than I added. The extra additions were either one liners, or new tests of the new directive. Had I thought to do the whole directive thing first, I would've handled adding a new idea in the last PR, which would've chunked it up a little better. But alas, lessons learned for the next feature :wink:

src/ideas/ideaInput/ideaInput.js
Mctalian: "since it's also a constructor"

Can you elaborate on this?

I will change the name (back to what I originally had it as), I just want to understand your comment completely.

src/ideas/ideaInput/ideaInputView.js
Mctalian: So if instead I added `ng-controller=...` to the tpl.html and left `controller: ...` out of the directive, it could be called a View? (albeit probably bad practice to do so, just trying to get my terminology down)

YashdalfTheGray:
All good.

src/ideas/ideaInput/ideaInput.js
Mctalian: Despite this being an outdated diff/comment, I'd still like to you to elaborate so I can learn.

src/ideas/ideaInput/ideaInputView.js
Mctalian: Despite this being an outdated diff/comment, I'd still like to you to elaborate so I can learn.

src/ideas/ideaInput/ideaInput.js
YashdalfTheGray: So the convention for Javascript is that for constructors, we use capitalized camelCase like `IdeaInputCtrl`. 

As to why it's a constructor, all controllers in Angular are created using Javascript constructor functions (for all intents and purposes) which we specify when we declare an `angular.module('blah').controller(function(() {  })`. In reality, they are factories that supply the constructor for when Angular instantiates a new instance of a controller but it's still a constructor function, so it's capitalized camel cased. 

In short, when you use `ng-controller` (ui-router uses this too), Angular asks for a constructor function of the name that is passed to `ng-controller` to new up an instance of the controller to attach to the `$scope` object for that view. It then injects a reference to the `$scope` object into your controller. If you pass the controller using the `controller as` syntax, a new instance is created and attached to the `$scope` for that view but a reference to that scope isn't passed to you. 

This is how I understand it but I don't have total knowledge about Angular's inner workings.

davethenipper:
I did not look over the tests much (coverage kinda proves that part), but the code looks good.  Approved by me! (Not sure how much weight that holds XD).

LongArmMcGee:
Same as dave

alanlgirard:
Everything looks good except if you go to edit an idea that has an event already assigned and remove that event by setting it to No Event, there is an error seen in the console window.

alanlgirard:
Approved! :+1:

Issue #254, opened 2016-01-25
Increase Title Character Limit


davethenipper:
The title needs a higher limit than what we have.

PR #263, opened 2016-01-27
Increase Idea Title Character Limit to 150

alanlgirard:
Increased the idea title character limit to 150 and fixed some over-flow/word wrapping issues

Closes issue  #254 

Simple review @YashdalfTheGray @Mctalian @monotkate @davethenipper

### DIFF
--- a/src/assets/app.css
+++ b/src/assets/app.css
@@ -304,13 +304,8 @@ h2.no-bottom-margin {
 }
 
 .word-wrap {
-    -ms-word-break: break-all;
-    word-break: break-all;
     word-break: break-word;
-    -ms-hyphens: auto;
-    -moz-hyphens: auto;
-    -webkit-hyphens: auto;
-    hyphens: auto;
+    word-wrap: break-word;
 }
 
 .identicon-home {
--- a/src/assets/app.css
+++ b/src/assets/app.css
@@ -298,16 +298,15 @@ h2.no-bottom-margin {
 }
 
 .ellipsis-overflow {
+    white-space: nowrap;
     overflow: hidden;
     text-overflow: ellipsis;
 }
 
 .word-wrap {
-    -ms-word-break: normal;
-    word-break: normal;
-    word-break: normal;
-    overflow-wrap: normal;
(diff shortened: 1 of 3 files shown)


src/assets/app.css
```diff
@@ -298,16 +298,15 @@ h2.no-bottom-margin {
...
 
 .word-wrap {
-    -ms-word-break: normal;
-    word-break: normal;
-    word-break: normal;
-    overflow-wrap: normal;
-    word-wrap: normal;
+    -ms-word-break: break-all;
+    word-break: break-all;
+    word-break: break-word;
```
Mctalian: So this causes some weird behavior in the description. Words like crazy get broken on the end of the line. I was the one that changed it to normal, where were we seeing undesirable wrapping?

src/ideas/ideaInput/ideaInput.tpl.html
```diff
@@ -2,12 +2,13 @@
     <div layout="column">
         <md-input-container>
             <label>Title</label>
-            <input ng-model="idea.title"
+            <textarea ng-model="idea.title"
```
Mctalian: It's big enough to need a text area now?! Haha, that might be too big. I just noticed 30 was always a lot shorter than I anticipated. I'm fine with whatever the other reviewers are fine with. I defer to better judgement.

alanlgirard: I switched it to a textarea for smaller resolutions where 150 characters will wrap.

Mctalian: :+1:

src/assets/app.css
alanlgirard: Didn't check IE.  The issue was if a title was 150 characters without a space it wasn't being broken up properly. That is why I put in the break-word. I will look into this tonight.  I hate having to do cross-browser support

YashdalfTheGray:
This seems good, approved!

Mctalian:
Approved. When I get into the office I'm going to pull all of this stuff down.

PR #267, opened 2016-01-29
Improving Code Coverage Reporting and CodeClimate foundations

Mctalian:
This PR adds code coverage output to the build console in the Travis builds so that we can see the coverage results for anyone's build without needing to post screen shots. There's also a nice at-a-glance view right after you run your tests now:

image

In addition, I've laid the groundwork for CodeClimate to be enabled for our repo. All that needs to happen now is @YashdalfTheGray creates a CodeClimate account and gets the CODECLIMATE_REPO_TOKEN and adds it as an Environment Variable to the Travis Build.

We might need to configure CodeClimate a bit more. We should probably ignore *.spec.js and *.mock.js and we might also need to tweak a little bit to get it to run for PRs. Coverage might only ever run for the default branch, not sure so that might be impossible. This gives us new badges we can add as well, but I must say CodeClimate is picky!


Mctalian:
@YashdalfTheGray (CodeClimate stuff)
@davethenipper (You like coverage)
@monotkate (You write a lot of tests and would probably appreciate coverage stuff)

monotkate:
ooooooOOOOOOOoooooooo

YashdalfTheGray:
Good stuff, I've already got a code climate account, once this is merged, I can pick it up from there.

davethenipper:
I LOVE Coverage!

Mctalian:
I'm assuming these are all approvals. @davethenipper confirmed his was.

Merging...

YashdalfTheGray:
yep, sure was.

PR #269, opened 2016-01-30
DialogBackCtrl Tests + Roles Requested Functionality

### LINKED ISSUE #220: Hardening Tests
Acceptance Criteria - all test files in the green
- [x] ideasView
- [ ] DialogBackCtrl

Following up on @Mctalian 's work, some of the tests need to be improved.  Working on ideasView at the moment.  It may be slowly because I'm still learning testing, but I want to do it.

monotkate:
- [x] hardening for DialogBackCtrl
- [x] roles requested added to ideas
- [x] fixed tests after roles requested broke them (I'm giving legit all the credit for that to @Mctalian )
- [x] added tests for roles requested

This one should be pretty self explanatory. @YashdalfTheGray @Mctalian @bcbrennecke (just because I'm adding commits and I thought I'd let you know Braun :smile: )

Closes #220


bcbrennecke:
Denied.  Too many commits.

monotkate:
Just the way I like it 😁

src/ideas/ideasView/ideaBack/dialogBack.spec.js
```diff
@@ -0,0 +1,134 @@
...
+            userSvc: userSvcMock
+        });
+    });
+
+    it('should exist', function() {
+        expect(ctrl).toBeDefined();
+    });
+
+    describe('$scope.cancel', function() {
+        it('should cancel', function() {
```
YashdalfTheGray: Can you make this say something else other than `$scope.doSomething does something`? Just making sure we're writing well documented code.

monotkate: Updated it.  Not sure what else to say since that's legit the only thing that function does is cancel the dialog window....

generateData.js
```diff
@@ -123,7 +123,8 @@ insertResult = db.ideas.insert(
                 time: ISODate(),
                 types: [{ name: 'Owner', _lowername: 'owner' }]
             }],
-            team: [{ _id: new ObjectId(), memberId: userIds[0] }]
+            team: [{ _id: new ObjectId(), memberId: userIds[0] }],
+            wantedBacks: []
```
Mctalian: Still have references of wantedBacks instead of rolesreq

src/ideas/ideaInput/ideaInput.spec.js
```diff
@@ -30,10 +37,6 @@ describe('ideaInput', function() {
             eventSvcMock = _eventSvcMock_;
             compile = $compile;
         });
-
-        var undef;
-
-        element = undef;
```
Mctalian: Why did we remove these? This was supposed to undefine element before each test. It's more of a clean slate kinda deal probably doesn't make any difference but, I'd prefer we kept it since there is a bit of logic that would make element change based on certain scenarios. This is probably the paranoid firmware engineer in me.

src/ideas/ideasView/ideaBack/dialogBack.spec.js
```diff
@@ -0,0 +1,134 @@
...
+            userSvc: userSvcMock
+        });
+    });
+
+    it('should exist', function() {
+        expect(ctrl).toBeDefined();
+    });
+
+    describe('$scope.cancel', function() {
+        it('should cancel the dialog', function() {
```
Mctalian: Idk if you can check this but you can say "cancel the dialog, reverting any changes" and also ensure that your data was reverted.

Could be more complicated than we need, though. Just a suggestion.

src/ideas/ideasView/ideasView.js
```diff
@@ -238,7 +239,12 @@ angular.module('flintAndSteel')
 
             ctrl.editIdea = function(idea) {
                 if (ctrl.isUserAuthor()) {
-                    ideaSvc.editIdea($scope.idea._id, idea.title, idea.description, idea.tags, [], idea.eventId).then(function() {
+                    var temp = [];
+                    _.forEach($scope.idea.rolesreq, function(roles) {
+                        temp.push({name: roles.name, _lowername: roles._lowername});
+                    });
+                    $scope.idea.rolesreq = temp;
```
Mctalian: What's this for? Why can't we just pass $scope.idea.rolesreq?

src/ideas/ideasView/ideaBack/dialogBack.spec.js
Mctalian: Just to be clear I'm even less crazy about my suggestion the more I think about it. If you want to try it go for it but the minute it gets annoying, I'd say leave it. Again I didn't look close enough at all, could be impossible.

Mctalian: I looked at the code. I'm pretty damn sure what I suggested is impossible. I recommend ignoring all of this. It literally just calls cancel on the dialog, as you mentioned :P

src/ideas/ideaInput/ideaInput.spec.js
monotkate: I think I just went over it with the new code you gave me. I can put it
back in

src/ideas/ideasView/ideasView.js
monotkate: It's legit for removing the $$hashkey because of how we store the chips in
the backend. We do the same thing for backs actually. It's dumb. This is
why I keep complaining about it.

src/ideas/ideasView/ideaBack/dialogBack.spec.js
monotkate: Yeaaaaaah that's why I've been like uh, guys? This is legit all it does....

Mctalian: I mean it *does* do more, but it's a weird case. The controller's scope is isolated so all of these changes occur to a local copy of the idea. Cancel closes the dialog and doesn't pass those changes back. So there's a key element there, but I think it's  impossible to test. If in addition to the `$mdDialog.cancel()` you were also resetting things back to their original values, there would be more to check, but it goes back to the isolated scope which is a benefit in this case.

So you were right, I say it's fine the way it is.

src/ideas/ideasView/ideasView.js
Mctalian: Hmm... yeah that's rough. I think for these types of things I use the `delete` keyword but this is fine. I figured there was some additional property that we needed to get rid of.

This is probably something we'll have to discuss further as a team to get a better understanding of the headaches and find a suitable solution.

generateData.js
Mctalian: Ok rolesreq is already here, so not "instead of" but "in addition to".

src/ideas/ideasView/ideasView.js
monotkate: Yeah, I'm totally a fan of changing how we work it to get rid of these dumb code snippets, but for now it is what it is.  

I switched to delete because it's cleaner looking code.

src/ideas/ideaInput/ideaInput.spec.js
monotkate: added it back in

generateData.js
monotkate: Yeah, thought I caught them all, but I missed that file, thanks for the check

src/ideas/ideasView/ideaBack/dialogBack.spec.js
monotkate: Yeah, I thought about whether or not there was a good way to see if the data had been changed or not.  With the controller the way it is though, it's like you said, I'm passing it an object, I'm cancelling the dialog, and therefore I am not passing the object back.  By the very nature of how it's written it can't really be tested.

server/ideas/ideas.js
```diff
@@ -47,7 +47,8 @@ module.exports = function(db) {
             likes: 1,
             backs: 1,
             tags: 1,
-            team: 1
+            team: 1,
+            rolesreq: 1
```
Mctalian: Do we need rolesreq in the idea headers? At this point I don't think we do. Sorry I missed this in my first pass.

server/ideas/ideas.js
```diff
@@ -70,7 +71,8 @@ module.exports = function(db) {
                             likes: doc.likes.length,
                             backs: doc.backs.length,
                             tags: doc.tags,
-                            team: doc.team.length
+                            team: doc.team.length,
+                            rolesreq: doc.rolesreq.length
```
Mctalian: Similar: Do we need rolesreq in the idea headers? At this point I don't think we do. Sorry I missed this in my first pass.

src/ideas/addIdeaView/addIdeaView.js
```diff
@@ -22,7 +23,7 @@ angular.module('flintAndSteel')
 
             $scope.addNewIdea = function addNewIdea(ideaToAdd) {
                 ideaToAdd.authorId = userSvc.getProperty('_id');
-                ideaToAdd.rolesreq = [];
+                //ideaToAdd.rolesreq = [];
```
Mctalian: You can probably just remove this altogether. It was a placeholder.

Mctalian:
I missed a few minor things. Once those are removed, this is approved!

server/ideas/ideas.js
monotkate: Oh no we probably don't. That was just me not understanding and looking for
tag definitions at some point. I can pull that tomorrow

monotkate:
Fixed.  Can I get a second approval?  @YashdalfTheGray  or @alanlgirard maybe?

YashdalfTheGray:
Yeah, this looks good, approved!

PR #272, opened 2016-02-01
Enable Markdown for Idea Description

### LINKED ISSUE #264: As a user, I want to be able to format the text description of my idea.
When describing my idea it is often necessary to format some text to better convey information. Typical formatting includes line returns to create rudimentary lists.

Mctalian:
I was bored... I'm sorry. And honestly the lack of newline support bugged me for a while. I was doing a nodeschool workshopper that required `marked` to be pulled in and I was like "huh, that was easy, I wonder if it'd be as easy for flintandsteel." Well it was.

This should close #264 

## Demo

**Before**:
image

**After**:
image

## Reviewers

@LongArmMcGee, @YashdalfTheGray, @davethenipper, @alanlgirard 

- [x] First Reviewer
- [x] Second Reviewer


monotkate:
Just a note looking at this, did you add any tips for what does what?  Not everyone's going to know how to format that without help, and then they'll be complaining about how they can't format text.....

Mctalian:
For now, I mainly want to address the line breaks. We can add something like this later:

image

Unless we don't want to support Markdown after all. Which I'm also fine with, this was just a super easy thing to implement.

Mctalian:
Reason being, I have yet to find suitable documentation for "vanilla" markdown. Another option is to configure out markdown for "Github Flavored Markdown" or *gfm* in which case we can just use the exact same link GitHub uses. It is possible to configure this with the library we are using. If that is the choice of the group, I can move forward with that and add the link that I mentioned in the previous comment.

YashdalfTheGray:
I want to go wit Github markdown since they have a lot of cool features that vanilla markdown doesn't have like tables, code blocks, language-specific highlighting, etc. So I kinda want to make that call right away.

YashdalfTheGray:
This is approved, pending the change to GFM.

alanlgirard:
I agree with Yash about going with the github flavor of markdown.  We should also enable markdown support for the updates and comments.

Mctalian:
I'll agree with Updates, not comments. I'm willing to be swayed, though.

YashdalfTheGray:
I don't agree with comments, that's a little too much but I think updates would be awesome. I don't want anyone leaving a large enough comment that that becomes necessary.

Mctalian:
Alright, so this PR has a bit more work, but I'm happy to do it. It's for the good of the order. Thank you for your suggestions and comments.

I need to enable GFM and also add a "Markdown Supported" link to the appropriate areas. Also make Updates Markdown capable.

YashdalfTheGray:
We can split updates being markdown capable to another story if you want. There is a different but related user case for that. This way, this PR isn't blocked up and we can have preliminary markdown support.

Mctalian:
I think it's just about a single-line change. I'll keep Updates in this one.

One new user story could be adding a "Preview" tab or option similar to GitHub for the future.

Mctalian:
Update using markdown...

image

Mctalian:
IdeaInput has markdown supported link.

image

Updates have markdown supported link

image

Mctalian:
Btw, code blocks seem to work without GFM enabled in this. Tables may or may not work with just GFM, we may have to add "tables: true" in the config, and syntax highlighting requires another library (highlight.js).

I never use tables, so I didn't think to check.

YashdalfTheGray:
Looks good, approved!

davethenipper:
This is beautiful.  Approved.

PR #278, opened 2016-02-01
Updated Branding to not include Innovation Challenge

bcbrennecke:
Updated about page branding (Deleted the term innovation challenge)
Updated favicon to be lightbulb. (I can't get this out of my browser's cache, but the file is def updated)
Changed top 5 ideas to be top 10 ideas.
Changed 'Top Ideas to Trending Ideas
Review this easy commit!

- [ ] 1st Reviewer

- [ ] 2nd Reviewer

@Mctalian @YashdalfTheGray @alanlgirard @monotkate @LongArmMcGee


bcbrennecke:
Someone let me know if their browser will load the favicon.ico.  Mine now just has a white box, but the file is correct.  I think its a caching thing on my end.

Mctalian:
Everything looks good but I don't think the favicon is correct.

Mctalian:
I think you need to use a tool like this: http://www.favicon-generator.org/

I've never actually been able to get favicons working if I tried to create them myself.

Also, you might as well also change the logo in README.md :stuck_out_tongue_closed_eyes:

bcbrennecke:
Okay @Mctalian I generated a new favicon with a transparent background, it totally works on my stuff now.  Also the readme picture will update when this merges in.  It pulls from the same image as the about page

src/about/about.tpl.html
```diff
@@ -1,5 +1,5 @@
 <div layout="column" layout-align="space-between center" ng-controller="AboutCtrl as about">
-    <img src="assets/InnovationChallengeLogo.png" width="30%" height="30%" class="img-padding">
+    <img src="assets/InnovationChallengeLogo.png" width="800px" height="475px" class="img-padding">
```
alanlgirard: Setting this to a default width and height probably isn't the best solution.  It will cause the lightbulb to appear really large on mobile as seen below

image

Mctalian:
Favicon works. I agree with Alan, fix that and you have my approval.

bcbrennecke:
Took a bit to figure it out, but it now resizes with the width of the screen.
image
image

bcbrennecke:
I'm going to call this good and merge since that was the only problem.

Mctalian:
Man, you gotta adhere to the checkboxes!

Mctalian:
I approve, even though it's too late!

bcbrennecke:
Y'all said you approve except this one thing!  Soo I merged!  Conditional approval is still approval!

Mctalian:
Seriously though, that's the point of the checks. Sometimes I take an initial pass, put some comments down and take a second pass later. Future PRs should wait for checks, unless you merge a PR and a bug is found immediately, then you should fix that- or similar exceptions.

bcbrennecke:
That's fair.  Will do

PR #285, opened 2016-02-03
Fix Idea Title in IE

alanlgirard:
The idea title was not displaying properly in IE as can be seen on production right now.  Just removed an unused flex.  Also added in a `margin: 0px` so that the the spacing is the same in Chrome and IE.

@Mctalian @monotkate @YashdalfTheGray @bcbrennecke 

- [x] - Reviewer 1
- [x] - Reviewer 2

### DIFF
--- a/src/assets/app.css
+++ b/src/assets/app.css
@@ -81,6 +81,7 @@ img.grayscale:hover {
 }
 
 h1.no-margins {
+    margin: 0px;
     margin-block-end: 0;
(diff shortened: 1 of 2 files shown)


monotkate:
meant to fix that, didn't see it breaking though.  Didn't check IE, my bad.  Thanks for the fix @alanlgirard

Mctalian:
Approved

PR #287, opened 2016-02-03
Disable events

### LINKED ISSUE #277: Hide Events
For now it looks like we do not want user's to tag events themselves.  Holly is on board with the idea of allowing engineers to work on things outside of Logix.

Mctalian:
This is kinda temporary since there's a bug and we need to discuss how events are going to be utilized in the future.

Closes #277 

@YashdalfTheGray, @davethenipper, @alanlgirard 

Reviewers
- [x] First
- [x] Second

### DIFF
--- a/server/routes.js
+++ b/server/routes.js
@@ -58,7 +58,7 @@ module.exports = function(app, db) {
             req.body.title,
             req.body.description,
             new ObjectId(req.body.authorId),
-            new ObjectId(req.body.eventId),
+            req.body.eventId,
             req.body.tags,
             req.body.rolesreq
         ).then(function(doc) {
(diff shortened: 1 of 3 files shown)


alanlgirard:
Approved.  We should log an issue to revisit the events so we don't forget about it.

bcbrennecke:
Approve.

Mctalian:
@alanlgirard #288 Created

LongArmMcGee:
My opinion on the matter: events are a useful tool to organize ideas according to real world activities. If a specific group doesn't want to use them that's fine. That doesn't mean someone else won't.

Mctalian:
Please see the new issue with discussion around events. Right now there is
a bug with events due to a change that was pushed so this is a temporary
disabling until we get a better idea of how to handle events. It's probably
like #288 or #289 or something.

PR #291, opened 2016-02-04
Navigating to a non-existent idea behaves appropriately.

Mctalian:
This used to work, but the back-end changes broke this.

When navigating to an idea that doesn't exist (someone deleted it and you're following a link to the deleted idea), it was supposed to reroute you back to the home state and toast you with "Sorry, that idea doesn't exist" or something to that effect.

What it was doing was just displaying a blank idea page. Now it's fixed.

Reviewers: 
- [x] @bcbrennecke (or other)
- [x] @alanlgirard (or other)

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -139,6 +139,7 @@ gulp.task('start:dev', ['test:client', 'inject', 'generate:data'], function() {
         script: 'server/server.js',
         env: { 'NODE_ENV': 'development' },
         'ignore': [
+            'coverage/*',
             'server/datastore/*',
             'src/*'
         ]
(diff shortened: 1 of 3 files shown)


bcbrennecke:
Convenient!

YashdalfTheGray:
I have fixed this in the past. Something else broke this. Specifically
remember fixing deleting ideas after @Mctalian pointed it out.

Yash Kulshrestha
yash.kulshrestha@gmail.com
(614)563-1163

Mctalian:
That was slightly different. That was SSE related, this is just a bogus ID.

We now return 404 and IDEA_NOT_FOUND instead of 200, so the logic needed to
be moved to the error handler since it's not a 2xx HTTP status.

YashdalfTheGray:
Approved.

alanlgirard:
Approved

PR #293, opened 2016-02-06
minor text change

LongArmMcGee:
Is time a skill? No...

### DIFF
--- a/src/ideas/ideaInput/ideaInput.tpl.html
+++ b/src/ideas/ideaInput/ideaInput.tpl.html
@@ -63,7 +63,7 @@
         </div>
         <div layout="row" flex layout-wrap>
             <md-input-container flex>
-                <p><b>Choose what types of skills your team is looking for:</b></p>
+                <p><b>Choose what types of skills and assets your team is looking for:</b></p>
                 <div layout="row" layout-wrap>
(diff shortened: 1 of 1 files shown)


YashdalfTheGray:
This should be on your fork not on the main repo.

Mctalian:
Also I don't know how much I like "assets".

Would replacing skills with contributions be better?

monotkate:
Honestly, I'd probably change it to resources rather than assets or contributions.  I'd replace skills at that point as well.  I threw in skills because I couldn't think of anything better at the time.

Before even going there though, there's a bunch of things that need to be fixed in this PR.
1) Close this PR (do NOT merge).  Get off the master and put it in your own branch like @YashdalfTheGray said.  If you need help with this, ask someone.
2) There's a secondary reference to skills in ideasView.  If you're going to change one, you have the change the other.
3) Tag reviewers if you want your PR to go through.  We have established a new procedure where you add checkboxes in your PR description and then tag at minimum two users.  When two have checked the boxes voicing their approval, you can merge it.  

Without following proper procedure, even for a seemingly small PR like this, we can't really merge the PR.  Adding exceptions ends up cascading and in a project like this order is necessary.  Thanks!

Mctalian:
Well put @monotkate,  I'm just going to close, then.

PR #297, opened 2016-02-09
Value Ratings

monotkate:
Hey guys, so.  Value ratings work.  I know we said we didn't necessarily want to push any of this in without MVP being complete, but I did want to at least get feedback.  With tests the added functionality is about 500 lines of code.  Also, I know there are some leftover comments.  Some of that is stuff I want to get back to, some of it I want to make sure I'm not using, etc etc.  Once I gather some opinions I will likely close the pull and finish up (not merge.... unless I should be merging?).  I'm not sure if people want me to drop it all at once once I've got value and complexity good to go or what.

Things that work: 
- [x] Value can be rated by user
- [x] Average updates as users change rating - can be seen with hover
- [x] Value cannot be rated if not logged in
- [x] If not logged in, stars are black and show average rating.

Things I intend to add yet:
- [ ] Half stars for average (not sure about for ratings)
- [ ] Potentially hover for rating
- [ ] Complexity
- [ ] Input on idea creation
- [ ] Better arrangement (ideas welcome)

Opinions I'm looking for:
@YashdalfTheGray @alanlgirard @Mctalian @bcbrennecke
- Where to display this
- How is ease of functionality
- Have you found any bugs.  For a while it would clear the stars when you edited a back for instance. 
- Do we like this? do we hate this?


monotkate:
Shit, tests are going to fail, fixing now

database-schema.json
```diff
@@ -14,6 +14,10 @@
         "title": "",
         "description": "",
         "authorId": "one user._id",
+        "avgValue": {
+            "stars": {"filled": true, "filled": true, "filled": true, "filled": true, "filled": false},
```
Mctalian: Is this supposed to be an array of objects with `filled` properties?

database-schema.json
```diff
@@ -59,6 +63,13 @@
                 "memberId": "idea.authorId"
             },
             "set of objects with user._id"
+        ],
+        "value": [
+            {
+                "stars": {"filled": true, "filled": true, "filled": true, "filled": true, "filled": false},
```
Mctalian: Same comment on line 18: Is this supposed to be an array of objects with filled properties?

database-schema.json
```diff
@@ -14,6 +14,10 @@
         "title": "",
         "description": "",
         "authorId": "one user._id",
+        "avgValue": {
```
Mctalian: Is this actually stored in the database? Or calculated via mongo aggregate?

database-schema.json
```diff
@@ -59,6 +63,13 @@
                 "memberId": "idea.authorId"
             },
             "set of objects with user._id"
+        ],
+        "value": [
+            {
+                "stars": {"filled": true, "filled": true, "filled": true, "filled": true, "filled": false},
+                "value": 4,
```
Mctalian: Can't this be calculated/inferred via `stars.filter((element) => element.filled === true).length;`? Do we need to store this in the db? Better question is should we store it there? I'm open to either way, just would like some context as to why you chose this path.

database-schema.json
monotkate: yeah, that's an array, I'll change it

database-schema.json
monotkate: no, my bad.  This is calculated via aggregate

server/ideas/ideaModel.js
```diff
@@ -24,12 +24,13 @@ function Idea(title, description, authorId, eventId, tags, rolesreq) {
         types: [{name: "Owner", _lowername: "owner"}]
     }];
     this.team = [{memberId: this.authorId}];
+    this.value = [];
 
     return this;
 }
 
-exports.create = function(title, description, authorId, eventId, tags, rolesreq) {
+exports.create = function(title, description, authorId, eventId, tags, rolesreq, value) {
```
Mctalian: Are we ever going to create an idea with a non-zero value? This is more of a design decision. Should an author rate their own idea on creation?

database-schema.json
monotkate: well.... I mean.... right now I'm actually assembling the stars from the value, instead of the other way around.  You have to update both any time it's changed regardless.  The number is what's used by aggregate though so I wouldn't zap it that way.

server/ideas/ideaModel.js
monotkate: Honestly I was going to add that (check PR description).  We don't have to require it, but I wanted to add it for idea creation.

server/replaceIds.js
```diff
@@ -177,6 +177,15 @@ module.exports = function(db) {
...
 
+        var ideaAvgValue = {};
+        if (data.avgValue.length === 0) {
+            ideaAvgValue.push(new Promise(function(resolve) {
+                resolve();
+            }));
+        }
+        else {
+            ideaAvgValue = data.avgValue;
+        }
```
Mctalian: Something doesn't seem write about this code block. Why do we care about average value in replaceIds? We might care about replacing the userIds in each individual value in the `value` array, but not the calculated average value.

monotkate: I can look back at this.  I think it was an issue of it was wiping it because it wasn't keeping it, but it may be something I backtracked on but didn't remove.  It may have had something to do with the tooltip not refreshing properly.  I'll look into it.

server/routes.js
```diff
@@ -176,7 +211,36 @@ module.exports = function(app, db) {
...
+            }
+
+            //load star values to update page
+            avgIdea[0].avgValue.stars = [];
+            for (var i = 0; i < 5; i++) {
+                avgIdea[0].avgValue.stars.push({ filled: i < avgIdea[0].avgValue.value });
+            }
+
+            //return updated idea
+            return avgIdea;
```
Mctalian: Ok... now this makes sense why SSEs were a problem.

Did you consider doing the aggregate calculation on the way out instead of the way in? For instance, someone requests a particular idea, mongo does the aggregation on the existing value array, and appends that onto the idea, much like `replaceIds` functions. We could rename that file `postProcessIdea` and do this aggregation stuff during the idea retrieval or `GET` calls rather than the `POST` and `PATCH`. This way, you wouldn't have to worry about race conditions with two people doing the same thing at the same time. Both would trigger an SSE update which triggers a new `GET` call for the idea.

database-schema.json
Mctalian: So in that case can we do the stars stuff on the client side then? Altering two fields at once in the database seems unnecessary.

server/replaceIds.js
Mctalian: So if SSEs are changed up a bit on the client side to refresh the idea instead of just taking the idea directly from the SSE, we should be able to put your aggregation stuff here into "replaceIds" and rename "replaceIds" to "ideaPostProcess" or something. That way the file would make more sense and it should make your life easier and reduce code duplication.

server/ideas/ideaModel.js
Mctalian: FWIW, authors don't have the option to like their idea during creation, and many of them do from what I've seen so far on production.

I think it makes *more* sense to have value/complexity in idea creation, but I also think an author will fill it out once their idea is posted if they choose to.

I'm leaning toward not having it during creation, but I'm not sold yet.

server/ideas/ideaModel.js
```diff
@@ -24,12 +24,13 @@ function Idea(title, description, authorId, eventId, tags, rolesreq) {
         types: [{name: "Owner", _lowername: "owner"}]
     }];
     this.team = [{memberId: this.authorId}];
+    this.value = [];
```
Mctalian: Just noting that we need to change this if we choose to allow value rating during idea creation. If not, this is acceptable.

database-schema.json
```diff
@@ -59,6 +63,13 @@
                 "memberId": "idea.authorId"
             },
             "set of objects with user._id"
+        ],
+        "value": [
```
YashdalfTheGray: should probably say `"values"` since it's an array.

database-schema.json
YashdalfTheGray: Do we have to store this in the database? Are you able to use the `"value"` field below to fill out the stars?

database-schema.json
YashdalfTheGray: Same question here, are you able to fill out the stars using the `"value"` field?

database-schema.json
YashdalfTheGray: I agree with Bob here even though the `PATCH /ideas/:id` endpoint is able to support multiple patches coming through and update them.

server/ideas/ideaModel.js
YashdalfTheGray: Liking an idea is an interaction that happens after you create an idea. It's like liking your own post on Facebook. If Facebook marked all of my posts as liked when I created them, it would get annoying and I would look like a pompous ass. I know it's not the same here but there is a parallel that can be drawn from it. 

Also, the author already sees value in creating an idea, they are always going to rate their idea's value at 5/5 unless they are modest to a fault. I think that the author is the only one who shouldn't get a rating for value. We shouldn't ask the person creating the idea what the value of the idea is; we should ask that person to put their idea of value down in the description and others to rate how valuable they think it's going to be. 

In addition to that, asking the author to estimate their own idea's value skews the dataset right off the bat.

server/replaceIds.js
YashdalfTheGray: Yeah, I like the file rename and adding the aggregate stuff here instead of anywhere else. Everything related to getting an idea goes through this file (including SSE).

server/routes.js
YashdalfTheGray: Yeah, I agree here too. Calculating something that the client needs on the way out probably the best way of doing it.

src/assets/app.css
```diff
@@ -1,5 +1,36 @@
...
+    list-style-type: none;
+    display: inline-block;
+    cursor: pointer;
+
+    &.filled {
+      color: #ffcc00;
+    }
+}
+/*
+  &.readonly .star.filled {
```
YashdalfTheGray: This comment can be removed if you're not planning on using the `readonly .star.filled` class.

src/ideas/ideaSvc/ideaSvc.js
```diff
@@ -19,6 +19,10 @@ angular.module('flintAndSteel')
                 return $http.get('/api/v1/ideas');
             };
 
+            // this.getAvg = function getAvg(ideaId, type) {
+            //     return $http.get('/api/v1/ideas/' + ideaId, type);
+            // };
```
YashdalfTheGray: Same here, I think you can remove this as well.

src/ideas/ideaSvc/ideaSvc.js
```diff
@@ -87,6 +91,17 @@ angular.module('flintAndSteel')
                 }
             };
 
+            this.editIdeaRating = function editIdeaRating(ideaId, value/*, complexity*/) {
+                if (ideaId !== 'mock_idea') {
+                    return $http.patch('/api/v1/ideas/' + ideaId,
```
YashdalfTheGray: There is a function in `ideaSvc` to interact with the node endpoints. You should use that and if you need to add it, you can do so. The convention that we've had (for better or for worse) is that we've put all the `$http` calls in `ideaSvc` anything else idea related in the app is a client of `ideaSvc`.

src/ideas/ideaSvc/ideaSvc.spec.js
```diff
@@ -329,4 +342,36 @@ describe('ideaSvc', function() {
         });
     });
 
+    describe('ideaSvc.editIdeaRating', function() {
+        var editIdeaHandler, smallDummyIdea, patchOperation;
+
+        beforeEach(function() {
+            editIdeaHandler = $httpBackend.whenPATCH('/api/v1/ideas/' + dummyIdea._id, patchOperation).respond(200, 'OK');
```
YashdalfTheGray: Using `ideaSvcMock` instead of `$http` here will make it so that you won't have to use `$httpbackend` here. You might have to use it when you test any new additions to `ideaSvc` but that code then becomes isolated and `ideasView` then depends on `ideaSvc` when in turn depends on `$http` cleaning up our dependency tree. Right now, `ideasView` depends on `ideaSvc` and `$http` as well as `ideaSvc` being dependent on `$http`.

Sorry, I'm not trying to be a pain here, just trying to justify why using `ideaSvc` and `ideaSvcMock` is a good idea.

src/ideas/ideaSvc/ideaSvc.spec.js
```diff
@@ -329,4 +342,36 @@ describe('ideaSvc', function() {
...
+                value: dummyIdea.value
+            };
+
+            patchOperation = [
+                { operation: 'modify', path: 'value', value: JSON.stringify(smallDummyIdea.value) }
+            ];
+        });
+
+        it('should edit an existing idea', function() {
+            $httpBackend.expectPATCH('/api/v1/ideas/' + dummyIdea._id, patchOperation).respond(200, 'OK');
```
YashdalfTheGray: Same thing with `ideaSvcMock` and `$http` here.

YashdalfTheGray:
This is cool. Couple of requests, can you put up some screenshots of what they look like? 

I think this could go on what I've started to call the interaction bar, the little bar between the main idea card and the card with comments, updates, backs or whatever else on it. I don't agree with having the author rate either the value or the complexity, definitely not the value. Not on idea creation, not on the `ideaView` page. They should use the description to justify what have in mind and have the others rate the value and the complexity. 

I also think you should try doing what @Mctalian suggested with renaming `replaceIds` to `ideaPostProcessing` and running an `aggregate()` there. I even think that that might clear up the SSE issues that you're having and `replaceIds` really is post-processing an idea before it goes out the door anyway. Maybe put the server side changes up first and we can approve that if you're concerned about the PR lines of code limit.

Mctalian:
And, @YashdalfTheGray is correct. If we move this to "replaceIds" (and subsequently rename that file to "ideaPostProcessing"), the SSE thing will get handled by default with no changes to SSEs either server-side or client-side. I took a closer look and SSEs run `replaceIds` on the idea before pushing it to the clients by default, so integrating into that will make it work everywhere!

So that's good news!

Mctalian:
I just pulled it down to see what it looked like in person.

1. I think this should go on near the other interactions.
2. Is there a way we can wrap the stars in a container and give that the md-tooltip? Each star having one seems a bit wonky. Or maybe just hovering over the word "value"? Or even just an integer after the stars? Maybe in smaller text:
```
The text is half the height of the stars |   Avg: 3.0
                                                            |   You: 3.0
```
That probably won't work, just throwing some ideas out here.

monotkate:
I'm going to close this for now

PR #299, opened 2016-02-12
auth tokens plus unjanked the ui some

### LINKED ISSUE #256: As a user, I want to have my data secured against unauthorized modification so that my authored data can be protected.
This story is to make our rest API more secure. Yash, are you able to fill this in?

Edit by Yash: I'm jacking this comment so that I can add a checklist. 

- [x] Generate a uuid on the server for each user
- [x] Store the uuid into the user object in mongo
- [x] Pass uuid from the client for anything uses the `POST`, `PATCH` and `DELETE` HTTP verbs
- [x] Keep all the `GET` endpoints unprotected to be able to read data
- [x] Unprotect the `POST /users/login` endpoint
- [ ] Protect the `GET /users/:id` endpoint. No one should be able to pull user data without authentication (This may be tricky and might require user-agent access tokens instead so this might only send required information with no access token).
- [ ] Front End changes

YashdalfTheGray:
### Auth tokens
Created a new module on the server side called `tokens.js`. When it loads, it runs through all the users in the database and assigns them a token property. It also returns a couple of methods, one called `generate()` and another called `authorize()`. 
- `generate(user)` takes an object with an `_id` property and an `email` property and generates a token.
- `authorize(db, id, token)` takes a pointer to the db, a user id and a token to see if they match.

This all happens in `routes.js`, there is express middleware (called `processAuthorization`) declared at the top that handles the authorization and either calls the next middleware in the chain or responds with a `401 Unauthorized`. 

Routes protected (a.k.a. need authorization)
- `POST /ideas`
- `PATCH /ideas/:id`
- `DELETE /ideas/:id`
- `PATCH /users/:id`
- `DELETE /users/:id`
- `POST /events`
- `PATCH /events/:id`
- `DELETE /events/:id`
- `POST /comments`
- `PATCH /comments/:id`
- `DELETE /comments/:id`

I can explain the data involved in authorization when we talk about this card.  

### Unjanking
In no particular order,
- Removed echoing `$window.navigator` and regexp search results to console.
- Tooltip direction appeared on top of the idea title, changed direction to bottom.
- Feedback button was too wide due to misuse of `md-icon-button`; put the class in the right place, looks better. 
- Vertically aligned the user menu (when logged in), It was 6px off-center. 
- Changed the size of the identicon (30px) to match the rest of the icons (24px) on the toolbar. 
- Tooltips for feedback button and user menu/login were off to the side so the user menu tooltip would cover up the feedback button. Changed to bottom and changed feedback button tooltip to match.
- Removed inline style because no. We have the technology; it's called CSS. 

Will close #256 after merged. 

@Mctalian, @monotkate, @alanlgirard, @bcbrennecke 

- [x] Someone needs to test this in prod and see if the auth tokens work. (@McTalian)
- [x] First Reviewer (@McTalian)
- [x] Second Reviewer @bcbrennecke
- [x] First Functional Reviewer (@McTalian)
- [ ] Second Functional Reviewer


src/homeView/homeView.js
```diff
@@ -40,14 +40,12 @@ angular.module('flintAndSteel')
             $scope.Events = [$scope.Event2, $scope.Event3, $scope.Event4, $scope.Event5];
 
             function getInternetExplorerVersion(navObj) {
-                console.log(navObj);
```
Mctalian: Hahaha I just did this exact same thing in a PR I'm about to toss up, I'll see if I can undo. There's a little more I did, but I'll leave this portion to you since your PR came up first.

Mctalian: Or not... GitHub ignored my attempt to do that... so I'll handle the merge conflicts once they... *emerge* :boom:

package.json
```diff
@@ -31,6 +31,7 @@
     "express": "4.13.0",
     "external-ip": "0.2.3",
     "ip": "0.3.3",
+    "sha.js": "^2.4.4",
```
Mctalian: I know you like carets :rabbit: but I prefer having repeatable builds and upgrading dependencies once we're ready to do so.

YashdalfTheGray: Yep, that's a good idea.

server/routes.js
```diff
@@ -149,7 +184,8 @@ module.exports = function(app, db) {
         }
     });
 
-    app.delete('/api/v1/ideas/:id', function(req, res) {
+    app.delete('/api/v1/ideas/:id', processAuthorization, function(req, res) {
+        console.log('id: ', req.params.id);
```
Mctalian: Left a console log in here

server/routes.js
```diff
@@ -236,7 +272,14 @@ module.exports = function(app, db) {
                                 return res.status(200).json(users.errResObj);
                             }
                             else {
-                                return res.status(200).json(responseObj);
+                                var token = tokens.generate({ _id: responseObj._id, email: responseObj.email });
+                                users.update(responseObj._id, 'token', token, function(tokenError) {
+                                    if (err) {
```
Mctalian: I think `err` should be `tokenError`

server/routes.js
```diff
@@ -236,7 +272,14 @@ module.exports = function(app, db) {
...
+                                        console.log(chalk.bgRed(tokenError));
+                                    }
+                                    responseObj.token = token;
+                                    return res.status(200).json(responseObj);
+                                });
```
Mctalian: So when there is an error, we just want to print it and keep going with everything? I need some context here to understand the reasoning for no `else` or no `return` within the `if.

server/users/users.js
```diff
@@ -80,6 +81,7 @@ module.exports = function(db) {
         return new Promise(function(resolve, reject) {
             db.findOneById(COLLECTION, id).then(function(doc) {
                 if (getEntireObject) {
+                    doc.token = undefined;
```
Mctalian: Does this work?

src/ideas/ideaSvc/ideaSvc.js
```diff
@@ -25,12 +35,21 @@ angular.module('flintAndSteel')
                         parentId: parentId,
                         text: text,
                         authorId: authorId
+                    },
+                    {
+                        headers: {
+                            'Authorization': getAuthorizationString()
+                        }
```
Mctalian: You can probably change the `getAuthorizationString` function return this whole `{ headers: ... }` object instead so we're only repeating one line instead of 4 or 5. Up to you, might look cleaner.

Mctalian:
Alright, pending my comments and maybe some conversation, I approve the code. Gotta pull it down on pre-prod at work for functional review.

server/routes.js
YashdalfTheGray: You're right, that should be under an `else`.

server/users/users.js
YashdalfTheGray: The `GET /users/:id` endpoint is unprotected since the website breaks if I try to lock this down. We would want to lock this down since getting users would then send out their auth tokens too. So I fetch the entire user object and set the token property to `undefined`. Whatever is `undefined`, express doesn't send out. So we get the user without their token.

src/ideas/ideaSvc/ideaSvc.js
YashdalfTheGray: I want that `getAuthorizationString()` function to run each time the headers object is assigned to make sure we've got the right auth string. Evaluation happens at assignment so if I just store the headers object, the auth string wouldn't get updated after the first call.

Mctalian: You're saying this wouldn't work?

```javascript
function getAuthHeader() {
    return {
        headers: {
            'Authorization': 'Bearer ' + userSvc.getProperty('_id') + ':' + userSvc.getProperty('token')
        }
    };
}
```

server/users/users.js
Mctalian: Ok, got it. We don't want to send the token back otherwise people can pretend to be someone else. :+1: I didn't expand things out to see where we were or figure out how we would get here, so thanks for the context.

src/ideas/ideaSvc/ideaSvc.js
YashdalfTheGray: Yeah, that should work. I'm sorry, I don't know where my train of thought was leading me there. I will say that I will return the headers object, not the config object since I want the code to read like I pass in a config object to `$http` and in that, there is a field called `headers` which I get from `getCustomHeaders()`. 

```javascript
function getCustomHeaders() {
    return {
        'Authorization': 'Bearer ' + 
            userSvc.getProperty('_id') + ':' + 
            userSvc.getProperty('token')
    };
}
```

And then use it like

```javascript
$http.post('/ideas', newIdea, {
    headers: getCustomHeaders()
});
```

Mctalian: Ok, cool. I'd say that's still an improvement. If you disagree, that's cool- you can keep it as is, but Seeing 'Authorization' every time seemed like we could consolidate. I understand the whole not including `header` in there.

bcbrennecke:
I think it looks okay, functional review will probably be the telltale. I'll give it a check

Mctalian:
Pulling this down now.

Mctalian:
Editing and Deleting backs do not appear to work.

The server complains with:

`Error: Argument passed in must be a single String of 12 bytes or a string of 24 hex characters`

Mctalian:
I can still spoof who is the author of an idea with this :sob: 

image

I used my own Authorization: Bearer token in the header, but his authorId in the POST request body.

Can we cross check these things on the server side?

Or can we take authorId out of the POST request altogether and extract that from their auth token? I think that would be best.

server/routes.js
```diff
@@ -18,7 +18,42 @@ module.exports = function(app, db) {
...
+        passport = require('passport'),
+        Tokens = require('./tokens');
+
+    var tokens = new Tokens();
+
+    function processAuthorization(req, res, next) {
+        if (req.headers.authorization) {
+            var authorizationType = req.headers.authorization.split(' ')[0];
+            var authorizationToken = req.headers.authorization.split(' ')[1];
+            var userId = authorizationToken.split(':')[0];
```
Mctalian: Here, set `res.locals.authorId` to `userId`. We can then use this in every single request to get the authorId out of the token.

server/routes.js
```diff
@@ -53,7 +88,7 @@ module.exports = function(app, db) {
             res.status(500).json(error);
         });
     });
-    app.post('/api/v1/ideas', function(req, res) {
+    app.post('/api/v1/ideas', processAuthorization, function(req, res) {
```
Mctalian: In this function change `req.body.authorId` to `res.locals.authorId` in the `ideas.create()` call. In this diff, the change occurs on line 95.

server/routes.js
```diff
@@ -339,7 +386,7 @@ module.exports = function(app, db) {
         });
     });
 
-    app.post('/api/v1/comments', function(req, res) {
+    app.post('/api/v1/comments', processAuthorization, function(req, res) {
         Promise.all([
             ideas.get(req.body.parentId),
             comments.create(req.body.parentId, req.body.text, req.body.authorId)
```
Mctalian: Same here, `req.body.authorId` can become `res.locals.authorId`

server/routes.js
```diff
@@ -245,7 +290,7 @@ module.exports = function(app, db) {
         }
     });
 
-    app.delete('/api/v1/users/:id', function(req, res) {
```
Mctalian: Can we just remove this endpoint? I think this has the potential to hurt us WAY more than it will ever help us.

server/routes.js
```diff
@@ -268,7 +314,7 @@ module.exports = function(app, db) {
         });
     });
 
-    app.patch('/api/v1/users/:id', function(req, res) {
```
Mctalian: Do we use this endpoint in our app right now? If not, I think we should also remove it until we use it or until we can secure it properly. This is really dangerous.

YashdalfTheGray:
I can remove the endpoints but what's this `res.locals.authorId` stuff about?

Mctalian:
Approved knowing that more work is to come in a later PR

PR #300, opened 2016-02-12
Toolbar coverage

Mctalian:
Improves `toolbar.js` coverage by pulling the DialogController for login out and putting that into its own file. Also wrote a test for that resulting file, and cleaned up that file a bit to simplify it a little.

This one is minor cleanup and not really important (especially for MVP) so I won't spend time making demos and stuff. Feel free to review/approve/whatever whenever you want.

## Reviewers
@monotkate (original author), @YashdalfTheGray, @alanlgirard
- [x] Code 1 @YashdalfTheGray 
- [x] Code 2 @bcbrennecke
- [x] Func 1 @alanlgirard
- [x] Func 2 @monotkate


YashdalfTheGray:
Looks good, approved.

Mctalian:
To assist with some reviewing:

image

The only functionality this affects is Logging In since I pulled the Dialog out and into its own file:

toolbar-coverage

The above shows the following:
1. Opening the Login dialog and clicking outside to close it.
2. Opening the Login dialog and clicking the "X" to close it.
3. Opening the Login dialog and Logging in successfully.

Also adding @bcbrennecke to this because I am greedy and want reviewers.

alanlgirard:
Logging in with the a username that is not a part of the users collection throws a error on the server.  This issue is probably only on dev but we we should handle that case.  Other than that things look good.

bcbrennecke:
Code looks good, I'll pull down tonight for a functionality check if noone else beats me to it

Mctalian:
@alanlgirard is this just in this PR or an existing issue on master? Does
it crash the server?

alanlgirard:
This does not crash the server.  This issue looks to be an existing issue on master.  Prod handles it properly it seems but dev doesn't toast with wrong credentials and throws the error.  I would just check the prod logs to make sure it doesn't error also.

Mctalian:
I saw when you logged on and there was no similar error around that time.

Speaking of which, we should definitely change that toast. We're not
storing people's credentials and I don't want people to freak out.

monotkate:
I'll checkoff on functional test.  Seems to be fine as far as I can tell.  Same thing as Alan said

monotkate:
nice testing btw :+1:

Issue #295, opened 2016-02-08
As a user I want to view a FAQ page so i don't ask the developers stupid questions


davethenipper:
- [ ] Should provide old people with answers to their questions
- [ ] Should have a seperate page

bcbrennecke:
Closed with Pull Req. 302

PR #302, opened 2016-02-13
Learn Page (FAQ)!

bcbrennecke:
Alright here's the learn page.  Right now I have 4 items as samples.  I will add more before the pull request goes through, I just figured someone will want me to change something first.
Here's what you get:

**New Toolbar format**- We need to have static navigation to parts of the website, putting links up here just makes sense.  The word cloud is probably going to disappear soon, so we need a new place for these links to live.  This looks kind of funny right now with the lightbulb just hovering in the center with lots of space, it will look much better the minute we get the new logo.

**Learn Page**- I added the learn page as our faq page.  It has text describing the action on one side and then a GIF on the other side showing how to do that action.  Pretty self explanatory and easy to replicate for as many features as we want.  This page will also look a little different when we get the new branding.  The lightbulb will change here too.  I created a new css file for this page which lives in the learn folder, and all of the gifs for this page live in this folder.  It makes sense to organize them this way so it's easy to find and change the gifs frequently as we are making changes to the UI.

I forgot to revert my changes from my other pull request before starting this, so those changes also live in here too.  I will close the other request.  They were:
-Changed word cloud light color to accommodate projectors
-Removed old HelpCard code.

Code Reviewer   
- [x] @YashdalfTheGray ?
- [x] @McTalian

Functional Reviewer
- [x] @YashdalfTheGray ?
- [x] @McTalian

This will close issue #295 

image

1su4b4oj2s


Mctalian:
First off, I like where this is going. Great work on this.

I know you said you'd be adding more/changing the actual "Learning Subjects" but I would start with a "Login" one and put in the description that this is required to complete nearly every action on the site. That way you don't have to show that one over and over again in all of the other interactions! Instead you can say "First, login" or "After you've successfully logged in" in the descriptions. Or just assume they saw the Login gif first, so they should probably do that...

Secondly, I know this is a placeholder thing right now to see what it can look like (and it looks awesome), but if we can figure out a way to provide that data from the controller, I think that would be awesome, much like you did on the word cloud.

Third, the toolbar links, not sure if it's a Mac thing but it's not showing a "pointer" for the cursor. I think we should try to follow what angular does https://angular.io, https://angularjs.org/ where there's some feedback on hover (color changes, highlight, pointer cursor, highlighted selected tab, etc.) I like "Learn", maybe we can use "Browse Ideas" instead of just "Ideas"... or just "Browse" with a Lightbulb icon. I like the icon approach in the second example, much like our Login.

Much more future idea: I think eventually we can have a "Learn" dropdown like the second example has so we can inform people, not only about the platform but also about how to participate in the innovation challenge, maybe how to use our REST API, perhaps info about how we interface with SAFe or another framework. Just a thought for the future, don't do anything with this now.

bcbrennecke:
Update: Yes, this is IE compatible

Mctalian:
Hahaha, I didn't go there yet, but I'm glad it is!

src/learn/learn.tpl.html
```diff
@@ -0,0 +1,47 @@
...
+    </div>
+    <div flex="35">
+        <h2>Search</h2>
+        Searching is how you can find an ideas that have been submitted across the company.
+        The search bar looks through an idea's title, content, tags, and backs.
+    </div>
+</div>
+<div layout="row" layout-align="space-around center" class="learn-objects">
+    <div flex="35">
+        <h2>Like on an Idea</h2>
```
Mctalian: "Like on an Idea", maybe just "Like" or "Like an Idea"

bcbrennecke:
I have updated the above toolbar to include md-buttons so the links highlight when you move over them.  I also removed the add idea button in the sidebar because I added it to the above toolbar.  When the user is not logged in, tooltip says Log In to Create, and it just redirects to the page with the toast saying you must login to create an idea.  
Having all of the actions up in one place is going to be nice and much less confusing for the user.

src/learn/learnItem/learnItem.spec.js
```diff
@@ -0,0 +1,42 @@
+/* global angular */
+/* global describe */
+/* global module */
+/* global beforeEach */
+/* global inject */
+/* global it */
+/* global expect */
+
+describe('ideaItem', function() {
```
Mctalian: learnItem

src/learn/learnItem/learnItem.spec.js
```diff
@@ -0,0 +1,42 @@
...
+    "use strict";
+
+    var DIRECTIVE_NAME = "idea-item";
```
Mctalian: learn-item

src/learn/learnItem/learnItem.spec.js
```diff
@@ -0,0 +1,42 @@
...
+        var element = angular.element('<' + DIRECTIVE_NAME + '></' + DIRECTIVE_NAME + '>');
+        var compiledElement = compile(element)(scope);
+        scope.$digest();
+        return compiledElement;
+    }
+
+    beforeEach(function() {
+        module('flintAndSteel');
+        module('ui.identicon');
+        module('ideas/ideaItem/ideaItem.tpl.html');
```
Mctalian: lean item related stuff

src/index.html
```diff
@@ -76,11 +78,25 @@
             <md-button hide-gt-md class="md-icon-button" aria-label="Menu" ng-click="showNav()">
                 <i class="material-icons icon-button-adjust">&#xE5D2;</i>
             </md-button>
-            <h2><span>
-                <img ui-sref="home" class="center-img nav-bar-pointer" src="assets/Innovation_Challenge_Bulb.png" width="25"/>
+            <h2><md-button>
```
YashdalfTheGray: Does this `md-button` need to be inside an `h2`?

src/learn/learn.tpl.html
```diff
@@ -0,0 +1,12 @@
+<md-content layout="column" layout-align="space-between center" ng-controller="LearnCtrl as learn">
+    <div style="width:15%">
+        <img src="assets/Lightbulb.png"   class="lightbulb">
+    </div>
+    <div class="title" layout="column" layout-align="center center">
+        <h1>Learn</h1>
+        <div>Learn how to interact with Innovate.RA</div>
+    </div>
+</md-content>
+<md-item ng-repeat="LI in LearnItems">
```
YashdalfTheGray: Caps make `LI` seem like a constant. Should be `li`.

Mctalian: Can we go with `learnItem` since `li` is a known html element? Just seems weird to me having it there.

YashdalfTheGray:
There is an assets folder that I'd rather have the gifs hang out in since all of our images live there. It's also a web development convention which makes them easier to find. And you should put all the CSS classes in `learn.css` into our `app.css` code if you're not expecting to add a lot more CSS to it. CSS should also technically live in the assets folder but I've been meaning to move it to the root (on the same level as `index.html`) for a while now. 

I know breaking up the CSS file makes sense for human reading but sending another file for a little bit of CSS is not as performant as sending one file. Really, having CSS spread around in files doesn't make sense unless we're using LESS or SASS. `homeView.css` was an exception since that file was so large and had a lot of styles in it. But since the word cloud's going away, I think that should only be a little bit of CSS that can also be put in `app.css`.

src/learn/learn.tpl.html
YashdalfTheGray: That's a good idea, go with `learnItem`.

bcbrennecke:
Okay.  I addressed all of @Mctalian and @YashdalfTheGray comments.  Moved app.css, blew away learn.css, moved the gifs to a folder under assets, updated the learn tests to actually test learnitem changed LI to LearnItem, removed that h2 tag (i wasn't sure why it was there but left it before, it wasn't needed)

bcbrennecke:
Changed line height to the full height of the toolbar.  Removed button from the innovate logo.

Mctalian:
Approved.

PR #305, opened 2016-02-17
Revert Push to Master

alanlgirard:
Permissions on the repo level were not set properly so a push was able to be performed.  This reverts the changes


alanlgirard:
Moderated by @YashdalfTheGray

Issue #309, opened 2016-02-18
Unresponsive page


YashdalfTheGray:
Um...I don't even know what to say here...maybe this is only on dev but,

Steps to reproduce -
1. Open up flintandsteel in a tab
2. Interact with it, do something on the page, login, click around, whatever.
3. Open up another flintandsteel tab
4. Click around on that new tab

Result: the new tab is totally unresponsive...until, the old tab is closed and then it responds in a flash and goes back to being frozen again. 

The only fix is to not open 2 tabs on the website or to close all of the tabs down and reopen just one. 

It's bizarre, I'm hoping it's a one off but all the code that I have is doing it.

Mctalian:
I also noticed this tonight. When you look at the "network" tab in a chrome dev window, does it show a few files as "pending"?

YashdalfTheGray:
I didn't check the networks tab but I could make a guess that it's waiting on something. That seems to fit the description of something that it would be waiting on.

Mctalian:
Can you check and see if you're also experiencing that?

davethenipper:
I've used two tabs on production before (about 2-3 weeks ago) and it was
working fine.  Not sure if this changed.

Mctalian:
Yeah, I think production was fine as well. We'll have to back track and pin
down which commit.

Happy "Go-Live" Week! :P

alanlgirard:
Just confirmed that this is currently not an issue for production as it is currently so thats good, but we need to track down the issue before we push the next update to prod

Mctalian:
This *might* only be an issue on Chrome? I forgot about this and pushed all the recent commits to prod, and now I'm trying to track this down.

Mctalian:
I've tracked it back to commit 4445803 being the last working version. The next commit is replacing gif files so I'm not really sure how that would cause anything like that.

Mctalian:
But at commit 4445803, opening 3 tabs causes the issue.

Chrome only allows 6 (I think) concurrent connections to the same server at a given time... so I think this has something to do with the number of files we're trying to send at once in conjunction with each tab having a few SSE connections as well.

So I think we always had this issue, but it's getting worse overtime. This is something we'll have to look at closely to find a solution.

Mctalian:
Ok, I figured out why this came out of nowhere.

[...]

PR #330, opened 2016-03-01
Patch 309 and sseSvc refactor

Mctalian:
Addresses #309. Whether or not we want to close that out is another story (socket.io or WebSockets might be the official closure?), but this does buy us time.

**Problem**
Opening 2 tabs on Chrome would cause us to use up the 6 open connections that Chrome allows to a single server at a given time. This was caused because we were instantiating `SidenavCtrl` 3 times in `index.html`.

**Solution**
 I removed both of the extra `SidenavCtrl` instantiations and slightly refactored to remove the "dependency" on that controller in one of the "extra" places. (I used a `ui-sref`instead of the `navTo` call).

Furthermore, I upgraded sseSvc to be closer to a pub/sub pattern. The server is doing the publishing, so we don't need to worry about that. But the `create` call was refactored to a `subscribe` function. This function doesn't create a new EventSource (i.e. connection to the server) unless it has to. So the sidenav always has the "newHeaders" event connected. If you open up "Browse" it used to open up another one, but now it just subscribes its callback to that already open event, and the sseSvc calls all of the subscribed callbacks with that single connection.

I tried to loop all of the SSEs under one endpoint, but it wasn't cooperating, so I gave up.

**Outcome**
Opening 2 tabs does not degrade performance. If you do not navigate to an individual idea page, you can open up to 5 tabs without noticing any performance issues. If you are on an idea page for each tab, you can only open 3 tabs before you will notice a lack of responsiveness.

**A Better Future Solution**
[Web Sockets](http://stackoverflow.com/a/23015553/4026797), in which case I'll probably pull sseSvc out of flintandsteel and make it its own module? I think it has its uses with some tweaks. The subscription pattern I'm using now makes a few assumptions, and our sse endpoints are a little wonky because I'm still trying to learn what is going on with them.

- [x] Code Review 1 @alanlgirard
- [x] Code Review 2 @monotkate

Gifs can be created upon request. It works on Chrome and IE.

### DIFF
--- a/src/ideas/ideaBrowseView/ideaBrowseView.js
+++ b/src/ideas/ideaBrowseView/ideaBrowseView.js
@@ -19,10 +19,10 @@ angular.module('flintAndSteel')
                 console.log(response);
             });
 
-            sseSvc.create("newHeaders", "/sse/ideas", setIdeaHeaders);
+            sseSvc.subscribe("newHeaders", "/sse/ideas", setIdeaHeaders);
 
             $scope.$on('$stateChangeStart', function() {
-                sseSvc.destroy();
+                sseSvc.unsubscribe("/sse/ideas", setIdeaHeaders);
             });
         }
     ]
--- a/src/ideas/ideasView/ideasView.js
+++ b/src/ideas/ideasView/ideasView.js
@@ -99,10 +99,10 @@ angular.module('flintAndSteel')
                 }
             }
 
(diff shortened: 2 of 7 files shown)


alanlgirard:
Code looks good.  Also pulled it down and functions how expected. Buys us a little more time to really fix the issue.

Approved :+1:

PR #332, opened 2016-03-02
Remove ideas js (finally)

Mctalian:
This file has been around far too long. It was a super old way to generate the database back when we were preparing to pitch for the MKE Innovation Challenge last year, nearly 9 months ago, if I remember correctly.

I think I'll just let the build pass and merge it tomorrow late morning unless someone gives me a reason not to.


YashdalfTheGray:
Merge it.

PR #341, opened 2016-03-23
updated package versions

YashdalfTheGray:
To get this to run, you are going to have to run a `git clean -xfd` in your project root and then run an `npm install`. Tested it out, I think everything works and nothing broke as a result of package updates. 

Also, travis will probably tell us if something went south anyway. 

@alanlgirard @monotkate @Mctalian


YashdalfTheGray:
Travis seems happy. I'm looking for quick approvals on this so that we can get this merged in since there isn't a lot changed here.

alanlgirard:
I'll pull this down today just to verify that things are still happy but everything looks ok.

monotkate:
Got it, looks good

PR #343, opened 2016-05-27
7-Announcements Carousel

monotkate:
So....... testing is lacking because I have zero idea how to actually test the controller for this one.  Other than that, it should work.


monotkate:
Definitely forgot to run syntax tests 👍

monotkate:
Tests are fixed now, but there's a test that's failing due to identicons I apparently commented out and forgot about 👍

karma.conf.js
```diff
@@ -29,6 +29,8 @@ module.exports = function(config) {
             'src/lib/angular-messages/angular-messages.js',
             'src/lib/angular-identicon/src/identiconDirective.js',
             'src/lib/angular-identicon/src/md5.js',
+            'src/lib/angular-slick-carousel/dist/angular-slick.js',
```
alanlgirard: You are not using these anymore with your fancy do the carousel yourself code. You can remove these.

src/ideas/ideaItem/ideaItem.spec.js
```diff
@@ -16,7 +16,7 @@ describe('ideaItem', function() {
     function getCompiledElement() {
         var element = angular.element('<' + DIRECTIVE_NAME + '></' + DIRECTIVE_NAME + '>');
         var compiledElement = compile(element)(scope);
-        scope.$digest();
+        //scope.$digest();
```
alanlgirard: Should uncomment this line after we fixed the identicon resolution issue

src/utilities/whirligig/whirlCtrl.js
```diff
@@ -0,0 +1,82 @@
...
+            // Check to make sure the document has finished loading the html
+            //   before processing it and taging with of the elems are shown
+            $scope.load = function load() {
+                if (document.readyState !== 'complete') {
+                    return;
+                }
+
+                $scope.control();
+            };
+            var tid = setInterval($scope.load, 100);
```
YashdalfTheGray: so instead of using `setInterval` here, use angular's `$interval` so that the changes happen within an angular context.

YashdalfTheGray: Also, does that interval time need to be that low? Not questioning your decisions, just asking.

src/utilities/whirligig/whirlCtrl.js
```diff
@@ -0,0 +1,82 @@
...
+
+            // Every 5 seconds, navigate forward one slide
+            $scope.increment = function increment() {
+                $scope.navigate(1);
+            };
+            var slideInt = window.setInterval($scope.increment, delayTime);
```
YashdalfTheGray: Here too, use `$interval`. The one change with interval is that it doesn't return a timerId, it returns a function that you call to cancel the timeout. So instead of 

```javascript
var timerId = window.setInterval(someFunction, 1000);
clearInterval(timerId);
```
you would do

```javascript
var cancelInterval = $interval(someFunction, 1000);
cancelInterval();
```

src/utilities/whirligig/whirlCtrl.js
monotkate: no, I copied it from somewhere on the internet in order to fix a problem I was having *shrugs*.  it worked

src/utilities/whirligig/whirlCtrl.js
```diff
@@ -0,0 +1,82 @@
...
+                    ctrl.current = ctrl.items[ctrl.counter];
+                    ctrl.current.classList.add('current');
+                    ctrl.current.classList.remove('not');
+
+                    // If navigate is used, restart slide interval
+                    clearInterval(slideInt);
+                    slideInt = window.setInterval($scope.increment, delayTime);
+                }
+            };
+        }
```
YashdalfTheGray: Once this `$scope` goes out of...well scope, you're still going to have your timers running and every time you create new instances of this `$scope`, you'll create more timers. You should clean up your timers using

```javascript
$scope.$on('$destroy', function() {
    // this assumes that you're using $interval and not window.setInterval
    clearSlideInterval();
    clearLoadInterval();
})

File: src/utilities/whirligig/whirlCtrl.js (hunk 3 above)
Comment (YashdalfTheGray) [inline src/utilities/whirligig/whirlCtrl.js]: Play with that timer and see if you can raise it to like a second. It all depends on what the purpose of `$scope.load` is.

File: src/utilities/whirligig/whirligig.tpl.html [hunk 6]
```diff
@@ -0,0 +1,18 @@
+<div class="whirligig-container" ng-controller="WhirlCtrl">
```
YashdalfTheGray: You should tell angular that you intend to use `WhirlCtrl` with your directive by removing the `ng-controller`attribute  from here and putting it in the directive declaration object.

src/utilities/whirligig/whilirigig.js
```diff
@@ -0,0 +1,18 @@
...
+    [
+        function() {
+            "use strict";
+
+            return {
+                restrict: 'E',
+                scope: {
+                    title: '@'
+                },
+                templateUrl: 'utilities/whirligig/whirligig.tpl.html'
```
YashdalfTheGray: Use `controller='WhirlCtrl'` here to tell angular that this directive has a controller that is associated with it.

src/utilities/whirligig/whirlCtrl.js
monotkate: Sooo.... it's legit just waiting for the page to load so it doesn't try to manipulate any of the html before it exists.  Also, I threw in an console.log, and it executes a whopping one time.  I'm going to leave it as is time wise.

src/utilities/whirligig/whirlCtrl.js
monotkate: actually, that's not correct.  In order to stop the interval, your code would actually have to be as follows:

```var cancelInterval = $interval(someFunction, 1000);
$interval.cancel(cancelInterval);```

monotkate:
Ok, so I just threw in a fix.  I was messing around with the carousel and getting it to load smoothly because I was unhappy with jank on page load.  Fixed it, and then realized we still had annoying jank where some elements were loading before other and making for an extremely messy load experience.  So I fixed it.  Similar to how I did for the carousel, it checks to make sure the page has loaded, and once it has shows everything.  Before it's finished loading, it just hides it.  It doesn't slow down the load so much, but it makes the transition much smoother.  What do people thing? @alanlgirard @YashdalfTheGray @davethenipper

davethenipper:
I like it.  I viewed it on Alan's machine and everything looks good.

PO approved.

alanlgirard:
Things look good.

Approved

monotkate:
Does this mean I can push it? @YashdalfTheGray @alanlgirard

YashdalfTheGray:
Yeah, you can. I was going to be so productive tonight but it didn't happen. I meant to reply to this as soon as I saw this email but I saw it in a half asleep state. Moving sucks. Anyway, you should be good to go.

PR #349, opened 2016-08-06
5- All Ideas on home page

bcbrennecke:
This is very close.  Added a 'More...' button to the bottom of the top-ideas directive which [will be] hidden or shown depending on where it is used.  When a user clicks the 'More...' button, it adds 1 row to the page.

Help 1: I can't figure out how to hide the more button.  Something about passing that boolean into the directive, no matter what I try won't hide it.  Help?
Help 2: I'm failing the test for top-ideas.spec.js, but i'm not sure why.  Anyone have some insight here seeing what I added?


YashdalfTheGray:
You've got a parse error.

src/homeView/homeView.tpl.html
```diff
@@ -4,9 +4,10 @@
     <whirligig class="margin-top"> </whirligig>
 
     <!-- Top Trending Ideas -->
-    <top-ideas title='Top Ideas' ideaslist='topIdeas' sort='"-likes"'/>
+    <top-ideas title='Top Ideas' ideaslist='topIdeas' sort='"-likes"' showMoreButton = 0/>
```
YashdalfTheGray: This should be `showMoreButton="true"`. HTML attributes don't usually have spaces in them and they use double quotes instead of single quotes.

bcbrennecke:
Closing pull request as this isn't how the UI should be built anyway.  Should link to other page.

PR #351, opened 2016-08-19
Email Subscription List

davethenipper:
This PR includes the addition of the email list feature to the account page and the DB.

There is more work needed to refresh the account page on change of "isSubscribed" (see Re-hydrate user object card).

To test this you will need to check the values in the database (and ignore that the switch does not update on the account page).

@YashdalfTheGray @bcbrennecke @alanlgirard 
I need one tester and at least 2 approvers.

### DIFF
--- a/server/routes.js
+++ b/server/routes.js
@@ -318,7 +318,7 @@ module.exports = function(app, db) {
             res.sendStatus(500);
         });
     });
-
+    */
     app.patch('/api/v1/users/:id', processAuthorization, function(req, res) {
         var promises = [];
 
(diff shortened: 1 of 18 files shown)


YashdalfTheGray:
<- You! Test this!

server/routes.js
```diff
@@ -318,7 +318,7 @@ module.exports = function(app, db) {
             res.sendStatus(500);
         });
     });
-
+    */
     app.patch('/api/v1/users/:id', processAuthorization, function(req, res) {
```
davethenipper: Alan - Should check that we can't modify a user and data without authorization.

davethenipper:
Closing and removing in favor of a MailChimp Link on their profile.
This removes this story, as well as the re-hydrate user object.

PR #357, opened 2016-10-05
Login dialog fix

YashdalfTheGray:
**NOTE**: This includes changes from #356 so that should be merged before this.

* (fix) login dialog input fields did not lay out correctly
* (fix) we were using `ng-cloak` which interfered with the dialog animation
* (fix) buttons can't be flex containers in IE 
* (fix) buttons in dialogs are supposed to be right aligned
* (fix) spacing between the input fields and the button look better now


src/users/loginView/loginView.tpl.html
```diff
@@ -10,7 +10,7 @@
             </div>
         </md-toolbar>
         <span class="sign-in-tip" flex>Use your RA-INT Credentials</span>
```
alanlgirard: Remove flex

alanlgirard: Remove flex

alanlgirard: Remove flex

src/users/loginView/loginView.tpl.html
```diff
@@ -25,13 +25,11 @@
                     <div ng-message="required">Required</div>
                 </div>
             </md-input-container>
-            <br></br>
-            <md-dialog-actions flex layout="column">
-                <md-button class="md-raised md-primary login-button" type="submit" ng-disabled="loginForm.$invalid" flex>
+            <md-dialog-actions flex>
```
alanlgirard: Remove flex

src/users/loginView/loginView.tpl.html
alanlgirard: I was hoping that would tag the lines.....

alanlgirard: Remove flex

src/users/loginView/loginView.tpl.html
```diff
@@ -10,7 +10,7 @@
             </div>
         </md-toolbar>
         <span class="sign-in-tip" flex>Use your RA-INT Credentials</span>
-        <md-dialog-content class="extra-padding-15 extra-padding-top-md">
+        <md-dialog-content class="extra-padding-15 extra-padding-top-md" layout="column">
             <md-input-container flex>
```
alanlgirard: Remove flex

src/users/loginView/loginView.tpl.html
alanlgirard: Remove flex

src/users/loginView/loginView.tpl.html
alanlgirard: Remove flex

alanlgirard:
IE 
image

Chrome
image

PR #360, opened 2016-10-07
5 - All Ideas on home page (Infinite Scroll)

monotkate:
Infinite Scroll should work on the home page, as well as a button that appears/disappears as you scroll.

Added one test, but I didn't actually add that much code for it, just time.


src/homeView/homeView.tpl.html
```diff
@@ -1,4 +1,5 @@
-<div ng-show="loaded">
+<!DOCTYPE HTML>
```
YashdalfTheGray: The doctype directive should be in `index.html`, is there a reason that you have this here?

src/index.html
```diff
@@ -166,7 +168,7 @@
     </md-toolbar>
 
     <md-content flex layout="row">
-        <md-content ui-view flex></md-content>
+        <md-content ui-view flex > </md-content>
```
YashdalfTheGray: Is there just extra whitespace here?

bower.json
```diff
@@ -33,6 +33,7 @@
     "jquery": "2.2.0",
     "lodash": "4.1.0",
     "moment": "2.10.3",
-    "angular-material": "1.1.1"
+    "angular-material": "1.1.1",
+    "nginfinitescroll": "1.3.0"
```
YashdalfTheGray: Maybe you did this already but did you consider [ngMaterial's Virtual Repeat](https://material.angularjs.org/latest/demo/virtualRepeat) and rule it out?

monotkate: Because everything I looked up about infinite scroll pointed to this and I didn't know virtual repeat existed

src/homeView/homeView.tpl.html
monotkate: Trying to fix a bug. I can probably remove it

alanlgirard:
Tested in both Chrome and IE.  Everything seems to be working properly

bower.json
YashdalfTheGray: I'm conflicted. On the one hand, ngMaterial's virtual repeat probably works well with material design and angular since it's pretty much a first party library but on the other, you've done this work using `ngInfiniteScroll`. I'm not too keen on having jQuery in the codebase either to be honest.

monotkate: Well, these are your options: 
1. You use this because it works, and figure it out later
2. You don't use this because it uses jquery, and figure it out later

I do not likely have time to try to switch this to ngMaterial this week because I have far too many commitments to get done (ie, 2.5 hrs of Algorithms lecture + quiz and program, 2 chapters on software engineering, and job applications for unemployment).

YashdalfTheGray: Alright, what would you say to merging this for now and prioritizing switching to virtual repeat at the next available opportunity?

monotkate: I mean, out of all the things that need to be worked on, this is something that works, and works well.  So I feel like it's not even that much of a priority to switch them.  The only reason there is jquery, is because ui-router fucks up how it's detecting scroll.  This is something that I wouldn't be shocked if you run into exactly the same issue for virtual repeat.  Just saying.

monotkate: But I think it would be dumb of you to reject this very functional PR because it doesn't use the exact resources you wanted it to.

davethenipper:
PO Approved!

bower.json
YashdalfTheGray: I have resisted getting jquery into the app before and just because angular uses jquery doesn't mean that it's fair game for us to use it. Also if you look at the the angular-material website, they have virtual repeat embedded in multiple places and they don't have to use anything but angular. But I digress.

